PR #19, opened 2018-10-17
Refactor dialogs

mikx98:


src/worklogs/components/WorkLogDateInput.js
```diff
@@ -40,4 +35,10 @@ class WorkLogDateInput extends React.Component<Props, State> {
   }
 }
 
-export default WorkLogDateInput
+export const ControlledWorkLogDateInput = WorkLogDateInput
```
Siegrift: I am not sure if this is controlled component. Controlled components are input elements whose value is managed by React. I would just wrap the 'WorkLogDateInput' in withState HOC.

src/worklogs/components/WorkLogInput.js
```diff
@@ -50,36 +48,46 @@ class WorkLogInput extends React.Component<Props, State> {
...
+      () => final,
+      // actual value changed, but this components input stays
+      () => this.props.setValue(this.getText)
+    )
   }
 
-  handleInputChange = e =>
-    this.setState({ value: e.target.value, historyIndex: null })
+  handleInputChange = e => {
+    // e doesn't persist, we need to copy the ref
```
Siegrift:
```suggestion
    // events do not persist, we need to copy the ref
```

src/worklogs/components/WorkLogInput.js
```diff
@@ -11,33 +12,30 @@ const INCREMENT_BY_KEY = {
...
-    historyIndex: null
-  }
-
+class WorkLogInput extends React.Component<Props> {
   handleKeyDown = e => {
-    const { hints } = this.props
-    const { historyIndex } = this.state
+    const { hints, historyIndex } = this.props
 
     if (!(e.key in INCREMENT_BY_KEY)) return
```
Siegrift: in keyword also looks for properties in ancestry chain. Might be better to use Object.hasOwnProperty

src/worklogs/components/WorkLogInput.js
```diff
@@ -89,6 +97,16 @@ class WorkLogInput extends React.Component<Props, State> {
...
-}))(WorkLogInput)
+export const ControlledWorkLogInput = compose(
+  withState('historyIndex', 'setHistoryIndex', null),
+  withState('input', 'setInput', ({ value }) => value),
+  connect((state, props) => ({
+    hints: workLogInputHintsSelector(state, props)
+  }))
+)(WorkLogInput)
+
+const UnControlledWorkLogInput = withState('value', 'setValue', '')(
```
Siegrift: Just wrap in HOC

src/worklogs/components/WorkLogInput.js
```diff
@@ -50,36 +48,46 @@ class WorkLogInput extends React.Component<Props, State> {
       .map(acc => Math.max(acc, 0))
```
Siegrift: Also I think you can simplify this part by writing:
acc === hints.length ? null : acc

src/worklogs/components/WorkLogInput.js
```diff
@@ -50,36 +48,46 @@ class WorkLogInput extends React.Component<Props, State> {
...
       .map(acc => (acc === hints.length ? null : acc))
 
-    this.setState({ historyIndex: final })
+    this.props.setHistoryIndex(
+      () => final,
+      // actual value changed, but this components input stays
```
Siegrift: Imo better comment: Update input value according to new history index.

src/worklogs/components/WorkLogInput.js
mikx98: Hmm, yes it generally means the same, but as there are two conditions I want to test & apply, I find it prettier to go through .map.

mikx98:
@Siegrift pls check once again :pray:

PR #22, opened 2018-10-17
Calendar and other polish

mikx98:


src/store/actions.js
```diff
@@ -35,65 +35,58 @@ export const createPoint = (description: string, timestamp: string) => ({
...
+export const deletePoints = (indexes: number[]) => ({
   type: 'Delete points',
-  payload: indeces,
+  payload: indexes,
   path: ['points'],
   reducer: (state, payload) => state.filter((t, i) => !payload.includes(i))
 })
 
-export const editDescription = (index: number, description: string) => ({
+const editDescription = (indexes: number[], description: string) => ({
```
Siegrift: I think more natural word is 'indices'

src/store/actions.js
```diff
@@ -145,15 +132,27 @@ export const moveSelection = (increment: number) => (dispatch, getState) => {
...
+  dispatch(updateValue(['cursor'], 0))
+  dispatch(updateValue(['selection'], []))
 }
 
-export const setJiraIssue = (issue: ?string) => (dispatch, getState) => {
-  const { selection, cursor } = getState()
-  dispatch(
-    setJiraIssueForPoints(selection.length ? selection : [cursor], issue)
-  )
+// action factory
```
Siegrift: Why is this comment useful?

src/store/actions.js
```diff
@@ -145,15 +132,27 @@ export const moveSelection = (increment: number) => (dispatch, getState) => {
...
+const makeSelectionAction = action => (...args) => (dispatch, getState) => {
```
Siegrift: :+1:

src/worklogs/calendar/WorkLogEvent.js
```diff
@@ -0,0 +1,26 @@
...
+  event: Event
+|}
+
+const WorkLogEvent = ({ event: { issue, title, cursor } }: Props) => {
+  const showIssue = issue || cursor
+
+  return (
+    <React.Fragment>
+      {showIssue && (
+        <span style={{ fontSize: '75%', color: 'black' }}>
```
Siegrift: I think the styles needs to be adjusted because the contrast between blue and black is not that good imo. Also the issue label is not centered within the label.

src/store/actions.js
mikx98: will fix in another PR

src/worklogs/calendar/WorkLogEvent.js
mikx98: design will be fixed in separate PR, this is just a little something to work until then

PR #23, opened 2018-10-17
Little bit of styling

mikx98:


src/worklogs/components/WorkLogInput.js
```diff
@@ -37,6 +37,8 @@ class WorkLogInput extends React.Component<Props> {
   handleKeyDown = e => {
     const { hints, historyIndex } = this.props
 
+    e.key === 'Enter' && this.handleSubmit(e)
+
     if (!(e.key in INCREMENT_BY_KEY)) return
```
Siegrift: I have an eslint error here. I see that you turned this type of rule ("curly": ["error", "all"]) explicitely on in .eslintrc. Is there a reason for it? We should probably stich to vacuum config.

src/worklogs/components/WorkLogInput.js
```diff
@@ -37,6 +37,8 @@ class WorkLogInput extends React.Component<Props> {
   handleKeyDown = e => {
     const { hints, historyIndex } = this.props
 
+    e.key === 'Enter' && this.handleSubmit(e)
```
Siegrift:
```suggestion
    if (e.key === 'Enter') {
       this.handleSubmit(e)
       return
    }
```

src/worklogs/components/WorkLogInput.js
mikx98: it was a wish from @jakuub. As there are just a few people on this project, why not?

src/worklogs/components/WorkLogInput.js
```diff
@@ -84,17 +89,15 @@ class WorkLogInput extends React.Component<Props> {
     const { value } = this.props
 
     return (
-      <form onSubmit={this.handleSubmit}>
```
Siegrift: :+1:

src/worklogs/components/WorkLogInput.js
Siegrift: Just asking, I personally don't like it visually...

PR #24, opened 2018-10-17
Custom time point creation from normal input

mikx98:


Siegrift:
I understood the behavior from the code, but there is no evidence about using ctrl+enter to create a time point. Is this your idea or the spec needs to be updated?

mikx98:
It's in the spec :). Look for shortcuts in Logging mode.

Siegrift:
In the spec its the command key, i guess that is a windows key on other keyboards right?

hleumas:
No, it is ctrl on Windows. MAC users are using most of shortcuts with CMD instead of CTRL.

mikx98:
Since #26 it's accepting both Mac's command key and Ctrl key.

PR #25, opened 2018-10-25
Server + Jira authorization

mikx98:


config.js
```diff
@@ -0,0 +1,24 @@
+const transenv = require('transenv').default
```
Siegrift: When the .env file is missing you should probably get an error informing about .env file not being present (now you get a long error message informing about missing env properties).

package.json
```diff
@@ -4,44 +4,72 @@
   "private": true,
   "dependencies": {
     "@material-ui/core": "^3.2.0",
+    "autoprefixer": "7",
+    "babel-cli": "^6.26.0",
+    "babel-core": "6",
+    "babel-loader": "7",
+    "babel-plugin-transform-decorators-legacy": "1",
```
Siegrift: Are you using this?

package.json
```diff
@@ -4,44 +4,72 @@
...
+    "babel-plugin-transform-react-constant-elements": "6",
```
Siegrift: Are you using this?

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["env", "stage-0"]
```
Siegrift: I think you are missing React here:
https://babeljs.io/docs/en/babel-preset-react

package.json
mikx98: vacuum config webpack

package.json
mikx98: yop

.babelrc
mikx98: hmm, I think that including it doesn't really do anything. It works fine, imo. Do you have any specific reason why it should be listed there?

config.js
mikx98: I think the error is pretty much readable:
```
Error: Missing env: ft_host
[0] Missing env: ft_port
[0] Missing env: ft_webpack_port
[0] Missing env: ft_session_secret
[0] Missing env: ft_jira_consumer_key
[0] Missing env: ft_jira_consumer_private_key
[0] Missing env: ft_jira_request_url
[0] Missing env: ft_jira_access_url
[0] Missing env: ft_jira_authorize_url
[0] Missing env: ft_jira_api_call_myself
```
When you encounter such an error, you will look for the reason such env variable is not present and you will probably start exactly at `.env` file and find out that it's not there at all.

mikx98: If you still want to fix it, I would say that more sense it makes to PR here: https://github.com/vacuumlabs/transenv

Siegrift: I think you are right... The error could be a little better, but the missing properties are probably fine...

.babelrc
Siegrift: It definitely should, but it probably gets merged with these presets:  https://github.com/vacuumlabs/webpack-config-vacuumlabs/blob/e98e21d98d8908cbe42169e1c18339b0e148d512/src/babelLoader.js#L10

server/api.js
```diff
@@ -0,0 +1,23 @@
...
+
+const router = express.Router()
+
+router.get(
+  '/user',
+  catchExceptions((req, res, next) => {
+    const { userId } = req.session
+    if (userId == null) {
+      res.status(405)
+      res.json({})
```
Siegrift: You can chain calls on 'res'

server/auth.js
```diff
@@ -0,0 +1,105 @@
...
+      req.session.oAuthRequestToken = oAuthToken
+      req.session.oAuthRequestTokenSecret = oAuthTokenSecret
+      res.redirect(
+        `${config.jira.authorizeUrl}${req.session.oAuthRequestToken}`
+      )
+    }
+  })
+})
+
+router.get('/callback', (req, res) => {
```
Siegrift: Why is the endpoint named callback?

server/frontend.js
```diff
@@ -0,0 +1,28 @@
+import express from 'express'
+import path from 'path'
+import config from '../config'
+
+const router = express.Router()
+
+router.use(express.static(path.join(`${__dirname}/../build`)))
```
Siegrift: 👍

server/main.js
```diff
@@ -0,0 +1,54 @@
...
+import frontend from './frontend'
+import { errorHandler } from './utils'
+
+const MemoryStore = require('memorystore')(session)
+
+const app = express()
+
+if (config.forceHttps) {
+  app.use((req, res, next) => {
+    if (!req.secure && req.get('x-forwarded-proto') !== 'https') {
```
Siegrift: I don't understand this part of the code that much. Can you point me where to learn more about it?

server/main.js
```diff
@@ -0,0 +1,54 @@
...
+      res.redirect(301, `https://${req.get('host')}${req.url}`)
+    } else {
+      res.setHeader('Strict-Transport-Security', 'max-age=31536000')
+      next()
+    }
+  })
+}
+
+app.use(compression({ threshold: 0 }))
```
Siegrift: Magic constants rule apply here: define NO_COMPRESSION = 0 for this. Btw. why are we not compressing?

server/main.js
```diff
@@ -0,0 +1,54 @@
...
+
+app.use(
+  session({
+    store: new MemoryStore({
+      checkPeriod: 3600 // prune expired entries every 1h
```
Siegrift: What does this do? Will I be logout after 1 hour?

src/Root.js
```diff
@@ -0,0 +1,74 @@
...
+  constructor(props) {
+    super(props)
+    // when hot-reloading code, this gets reimported and the component is recreated.
+    // therefore we always start with no error
+    this.state = {}
+  }
+
+  // eslint-disable-next-line handle-callback-err
+  componentDidCatch = (error, info) => {
+    this.setState({ hasError: true })
```
Siegrift: Maybe save the error info and display it in the Houston component

src/Root.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  render() {
+    if (this.state.hasError) {
+      return <h1> Houston, we have a problem </h1>
```
Siegrift: 😄

src/Root.js
```diff
@@ -0,0 +1,74 @@
...
+  typography: {
+    useNextVariants: true
+  }
+})
+
+const Root = ({ store }) => (
+  <ErrorBoundary>
+    <Provider store={store}>
+      <BrowserRouter basename={process.env.REACT_APP_BASENAME}>
+        <DispatchProvider dispatch={store.dispatch}>
```
Siegrift: Why does DispatchProvider has to be here?

src/index.js
```diff
@@ -1,44 +1,27 @@
...
-    <BrowserRouter basename={process.env.REACT_APP_BASENAME}>
-      <DispatchProvider dispatch={store.dispatch}>
-        <App />
-      </DispatchProvider>
-    </BrowserRouter>
-  </Provider>,
-  document.getElementById('root')
-)
-registerServiceWorker()
+// registerServiceWorker()
```
Siegrift: I don't like code comments. If you want to disable it, I would prefer turning it off behind a flag and writing comment WHY are we disabling it.

src/login/Login.js
```diff
@@ -0,0 +1,61 @@
...
+        )}
+      </Button>
+    </Grid>
+  </Grid>
+)
+
+export default compose(
+  withStyles(styles),
+  withState('loading', 'setLoading', false),
+  withHandlers({
```
Siegrift: I believe you can just call setLoading directly and remove withHandlers HOC.

src/App.js
```diff
@@ -1,31 +1,29 @@
...
 
-const App = () => (
-  <MuiThemeProvider theme={theme}>
-    <Switch>
-      <Route path="/" exact component={WorkLogs} />
-    </Switch>
-  </MuiThemeProvider>
-)
+export default compose(
+  withState('user', 'setUser', null),
```
Siegrift: Why do we remember it only in local state? This could be stored in redux.

.babelrc
mikx98: yop, react is default babel preset and doesn't have to be listed. but that `@babel/preset-react` preset you are pointing to is something I am not sure of

server/auth.js
mikx98: because the oauth callbacks to this, when you click either Allow or Deny

src/Root.js
mikx98: it's because of pinto's data provider library - it's not used, but maybe in the future :thinking:

mikx98: it was there in the initial boilerplate

server/auth.js
Siegrift: I see, thanks!

server/main.js
mikx98: not really, it's used in both ems and pay overview apps with description that in production it should force https (you enable force_https env variable)

.babelrc
Siegrift: I think those are the same. If you go to https://github.com/vacuumlabs/webpack-config-vacuumlabs/blob/e98e21d98d8908cbe42169e1c18339b0e148d512/package.json there is only one babel preset react and it's this one. 

So in conclusion: This plugin has to be here because it's a peer dependency of vacuum webpack config, but doesn't  have to be listed (because it's already listed in VL webpack config)

server/main.js
Siegrift: I see, I would put a WHAT comment here (because the logic is quite hard to understand)

mikx98: I think the `if (config.forceHttps)` pretty much stands for itself. In .env there's also comment to enable it in production. doesn't it?

server/main.js
mikx98: hm, I think the `threshold: 0` stands for *force compress every time* .. but I think we should stick to defaults unless we have a reason for changing it (by default the threshold is 1kb) https://github.com/expressjs/compression

server/main.js
Siegrift: Yes there is. I think that's enough.

server/main.js
Siegrift: I see, thanks

mikx98:
:tada:

PR #26, opened 2018-10-27
Spec rewrite, key bindings refactor

mikx98:


src/worklogs/WorkLogs.js
```diff
@@ -78,78 +78,54 @@ export default compose(
...
+      const metaKeyPath =
+        ((e.ctrlKey || e.metaKey) && 'withCtrlKey') ||
+        (e.shiftKey && 'withShiftKey') ||
+        'plainKey'
+      const path = [mode, metaKeyPath, e.key]
+
+      const handler = get(keyBindings, path)
+      if (handler) {
+        handler()
+        // no side effects
```
Siegrift: I would remove this comment.

src/worklogs/WorkLogs.js
```diff
@@ -78,78 +78,54 @@ export default compose(
...
+            n: () => setJiraIssueForSelection(null),
+            ArrowUp: () => moveCursor(-1),
+            ArrowDown: () => moveCursor(1),
+            k: () => moveCursor(-1),
+            j: () => moveCursor(1),
+            Backspace: deleteSelection,
+            Delete: deleteSelection,
+            ' ': toggleSelectCursorPoint
+          },
+          withCtrlKey: {
```
Siegrift: Rename to withCtrlOrCommandKey

src/worklogs/WorkLogs.js
```diff
@@ -78,78 +78,54 @@ export default compose(
...
-          }
-          break
-
-        default:
-          break
+    })
+  ),
+  withHandlers({
+    handleKeyDown: ({ mode, keyBindings }) => e => {
+      const metaKeyPath =
```
Siegrift: I would put this into a simple helper function.
```suggestion
      const metaKeyPath = getMetaKeyPath(e)
```

src/worklogs/WorkLogs.js
```diff
@@ -67,9 +68,8 @@ export default compose(
       setJiraIssueForSelection: _setJiraIssueForSelection
     }
   ),
-  withHandlers({
-    handleKeyDown: ({
-      mode,
+  withProps(
```
Siegrift: You are not flowing the properties added by recompose HOC. I think the keyBindings object format should be enforced.

mikx98: lodash's `get` by path is not supporting flow and the component doesn't use it. So there's no really a place where it would make some good.

src/worklogs/WorkLogs.js
```diff
@@ -22,20 +23,20 @@ import {
...
 
-import type { Mode, DialogType } from '../consts'
+import type { Mode } from '../consts'
+
+const getMetaKeyPath = (
+  ctrlKey: boolean,
+  metaKey: boolean,
+  shiftKey: boolean
+) =>
+  ((ctrlKey || metaKey) && 'withCtrlKey') ||
```
Siegrift: If you write it using if/elseif/else you will save some letters and simplify the function.

mikx98: I think this is both more readable and shorter way to write it. I don't want really if-else conditions when they are not necessary

mikx98: there's curly eslint rule, so that suggestion wont be possible :(

Siegrift: :( Why do we have this rule set, @jakuub?

spec/src/index.adoc
```diff
@@ -11,12 +11,12 @@ Point:: Point in time. Two adjacent points mark an interval.
 Interval:: Marked by two adjacent points. Intervals can have associated work-logs and
 JIRA issues.
 
-Cursor:: User can use movement keys to navigate through intervals or points.
-Current interval/point is marked by cursor.
+Cursor:: User can use movement keys to navigate through intervals.
+Current interval is marked by cursor.
 
-Selection:: User can select multiple intervals (or points) to perform batch
+Selection:: User can select multiple intervals to perform batch
```
hleumas: What is the reasoning behind this change?

hleumas: Can you achieve deleting or changing the last logged point in this workflow?

mikx98: The reason is that points and intervals are really kind of *the same*. In the app user will see either points or intervals and imo intervals make more sense, because it's the representation that visualizes effort. It has duration and description. The point simply describes the following interval.

The last point, or better the **work in progress interval** is displayed in a special way, to indicate that it's not finished yet. But it can be edited the same way as other intervals.

src/worklogs/WorkLogs.js
mikx98: curly removed, suggestion applied

PR #27, opened 2018-10-28
User header + Vaccumlabs branding

mikx98:


src/assets/logo.svg
```diff
@@ -0,0 +1,13 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 338 64" width="250" height="64">
```
Siegrift: The logo would look much better if it was an svg instead of an image. Do we have it in svg as well?

src/common/Header.js
```diff
@@ -0,0 +1,66 @@
...
+            src={user.avatarUrl}
+            className={classes.item}
+          />
+        ) : (
+          <Avatar className={classes.item}>
+            <AccountCircle />
+          </Avatar>
+        )}
+
+        <Button
```
Siegrift: I would like this to look like in gmail... When you click the avatar, there would be a modal/dropdown opened, where you could logout or go to some account settings later on.

Anyway, this is also fine for now.

src/App.js
```diff
@@ -1,15 +1,29 @@
...
+import favicon from './assets/favicon.png'
+
 type Props = {|
   user: any
 |}
 
-const App = ({ user }: Props) => (user ? <WorkLogs /> : <Login />)
+const App = ({ user }: Props) => (
+  <React.Fragment>
+    <Helmet>
```
Siegrift: Can you please explain what the <Helmet> does? I am unable to understand it from the helmet README

Siegrift: I see now. Why not change the icon directly in index.html?

src/common/Header.js
mikx98: As there is only one option (to log out) right now, it is probably overkill. But I can definitely imagine it in the future (maybe when key bindings will be released).

server/frontend.js
```diff
@@ -6,13 +6,17 @@ const router = express.Router()
...
+
 router.get('/', (req, res, next) => {
   const appScriptSrc = `http://${req.hostname}:${config.webpackPort}/app.js`
+  const faviconSrc = `http://${req.hostname}:${config.port}/favicon.png`
 
   const html = `<!DOCTYPE html>
     <html lang="en">
       <head>
         <title>Fine Tracker</title>
+        <link rel="icon" type="image/png" href="${faviconSrc}" size="31x31">
```
Siegrift: :+1: I missed this one before, gj.

src/common/Header.js
```diff
@@ -0,0 +1,66 @@
...
+  logoContainer: {
+    flexGrow: 1
+  },
+  item: {
+    marginLeft: '1em'
+  }
+})
+
+const Header = ({ user, classes }) =>
+  user ? (
```
Siegrift: In this case I would use **branch** from recompose as this component is quite complex.

PR #28, opened 2018-10-30
Joining intervals

mikx98:


src/store/actions.js
```diff
@@ -90,6 +90,24 @@ const setJiraIssue = (indexes: number[], issue: ?string) => ({
...
+const joinIntervals = (indexes: number[], description: string) => ({
+  type: 'Join intervals',
+  payload: { indexes, description },
+  path: ['points'],
+  reducer: (state, payload) => {
+    const filterIndexes = payload.indexes.slice(1)
+
+    return state
+      .map(
+        (t, i) =>
```
Siegrift: It is really not easily understandable what is t, when the variable is called state. Please change to point. Also a good change might be not to call it a state but points.

src/worklogs/components/workLogInputHintsSelector.js
```diff
@@ -10,15 +12,53 @@ type Props = {
...
+const dialogSelector = (state: State) => state.modal
+const selectionSelector = (state: State) => state.selection
 const inputSelector = (state: State, props: Props) => props.input
 
+const showHintsFromSelectionSelector: InputSelector<
+  State,
+  any,
+  boolean
+> = createSelector([dialogSelector], dialog =>
+  Boolean(dialog && dialog.type === MODAL_TYPES.JOIN_INTERVALS)
```
Siegrift: This will always be boolean, remove Boolean wrapping.

src/worklogs/components/workLogInputHintsSelector.js
```diff
@@ -10,15 +12,53 @@ type Props = {
...
+)
+
+const pointDescriptionsSelector: InputSelector<
+  State,
+  any,
+  string[]
+> = createSelector([pointsSelector], points => points.map(t => t.description))
```
Siegrift: Why t? It's a point, so what you map through is 'point' or 'p'.

src/worklogs/components/workLogInputHintsSelector.js
```diff
@@ -10,15 +12,53 @@ type Props = {
...
+
+const selectionDescriptionsSelector: InputSelector<
+  State,
+  any,
+  string[]
+> = createSelector(
+  [pointDescriptionsSelector, selectionSelector],
+  (descriptions, selection) =>
+    descriptions.map((t, i) => selection.includes(i) && t).filter(Boolean)
```
Siegrift: Change the t name.

src/worklogs/components/workLogInputHintsSelector.js
```diff
@@ -10,15 +12,53 @@ type Props = {
...
+)
+
 const distinctDescriptionsSelector: InputSelector<
   State,
-  Props,
+  any,
```
Siegrift: Are we really returning any?

src/worklogs/components/workLogInputHintsSelector.js
```diff
@@ -10,15 +12,53 @@ type Props = {
...
-  // in order to keep the last occurence (the most recent)
-  reverse(uniq(reverse(points.map(point => point.description))))
+> = createSelector(
+  [
+    pointDescriptionsSelector,
+    selectionDescriptionsSelector,
+    showHintsFromSelectionSelector
+  ],
+  (allDescriptions, selectionDescriptions, showHintsFromSelection) =>
+    // in order to keep the last occurence (the most recent)
```
Siegrift: This comment should be moved on line bellow.

src/worklogs/components/workLogInputHintsSelector.js
mikx98: nope

src/worklogs/components/workLogInputHintsSelector.js
mikx98: that's not the 3rd (return) type, that's the props type... and the function doesn't depend on props, so it can be anything

src/worklogs/components/workLogInputHintsSelector.js
mikx98: nope ;)

PR #29, opened 2018-11-02
Get issues from Jira API

mikx98:


.env.template
```diff
@@ -23,4 +23,5 @@ ft_jira_request_url=https://vacuum.atlassian.net/plugins/servlet/oauth/request-t
 ft_jira_access_url=https://vacuum.atlassian.net/plugins/servlet/oauth/access-token
 ft_jira_authorize_url=https://vacuum.atlassian.net/plugins/servlet/oauth/authorize?oauth_token=
 
-ft_jira_api_call_myself=https://vacuum.atlassian.net/rest/api/2/myself?expand=groups
+ft_jira_api_call_myself=https://vacuum.atlassian.net/rest/api/3/myself?expand=groups
+ft_jira_api_call_issues=https://vacuum.atlassian.net/rest/api/3/search?fields=key&maxResults=100&startAt=
```
Siegrift: Put \n so that when you add a new line to this file it won't be shown as change in git.

Siegrift: I also suggest for you this plugin: https://marketplace.visualstudio.com/items?itemName=riccardoNovaglia.missinglineendoffile

config.js
```diff
@@ -18,7 +18,8 @@ module.exports = transenv()(({ str, bool }) => {
       requestUrl: str('ft_jira_request_url'),
       accessUrl: str('ft_jira_access_url'),
       authorizeUrl: str('ft_jira_authorize_url'),
-      apiCallMyself: str('ft_jira_api_call_myself')
+      apiCallMyself: str('ft_jira_api_call_myself'),
+      apiCallIssues: str('ft_jira_api_call_issues')
```
Siegrift: Trailing comma (if not forbidden by eslint)

server/api.js
```diff
@@ -20,4 +22,49 @@ router.get(
   })
 )
 
+router.get(
+  '/issues',
+  catchExceptions(async (req, res) => {
+    const getHandler = (resolve, reject) => (error, data) => {
+      if (error) {
+        reject(JSON.stringify(error))
+        res.status(405).json({})
```
Siegrift: I don't know what is the best practice for throwing HTTP status codes, but 405 (Method not allowed) is not a good status  imo. I would just return general 400. What do you think? Do you know how to use the status codes? :smile:

server/api.js
```diff
@@ -20,4 +22,49 @@ router.get(
...
+      } else {
+        const dataObj = JSON.parse(data)
+        const total = dataObj.total
+        const newIssues = dataObj.issues.map(t => t.key)
```
Siegrift: Why do you write the map argument as 't'? What does 't' stand for? I would prefer 'v', 'val', or 'value'

server/api.js
```diff
@@ -20,4 +22,49 @@ router.get(
...
+        resolve({ newIssues, total })
+      }
+    }
+
+    const fetchIssues = startAt =>
```
Siegrift: This function is fetching only up to next 100 issues. The function name should reflect that (e.g. **fetchNextIssueBatch**)

server/api.js
```diff
@@ -20,4 +22,49 @@ router.get(
...
+    let issues = []
+    let startAt = 0
+    let issuesLeft = 0
+    // eslint-disable-next-line no-constant-condition
+    do {
+      const { newIssues, total } = await fetchIssues(startAt)
+      issues = [...issues, ...newIssues]
+      issuesLeft = total - issues.length
+
+      startAt += 100
```
Siegrift: Make 100 a constant, (e.g. MAX_JIRA_ISSUES_IN_BATCH)

src/store/actions.js
```diff
@@ -156,3 +156,12 @@ export const deleteSelection = makeSelectionAction(deletePoints)
 export const editDescriptionForSelection = makeSelectionAction(editDescription)
 export const editTimeForSelection = makeSelectionAction(editTime)
 export const setJiraIssueForSelection = makeSelectionAction(setJiraIssue)
+
+export const getIssues = () => async (dispatch, getState) => {
+  // we already got the issues
```
Siegrift: Don't comment WHAT the code does (I can read that from the next line), better to write WHY or HOW.

Btw. how do we update the issues now? I can imagine new issues being created, thus we need to refetch them.

src/store/actions.js
```diff
@@ -156,3 +156,12 @@ export const deleteSelection = makeSelectionAction(deletePoints)
...
+  if (getState().issues.length) {
```
Siegrift: Eh, those curly :cry:

.env.template
mikx98: it's there already, it just wasnt :)

config.js
mikx98: it is forbidden

src/store/actions.js
mikx98: :cry:

src/store/actions.js
mikx98: by reloading the app ... it probably doesn't happen that often

server/api.js
mikx98: i went by this https://sk.wikipedia.org/wiki/Stavov%C3%A9_k%C3%B3dy_HTTP#4xx_Client_Error, please check once again :pray:

mikx98:
:tada:

Issue #33, opened 2018-11-02
Help dialog on ?


mikx98:
Show either converted spec or some key shortcuts.

jakuub:
I will reopen this, because I reallized, that each time, I don't remember the shortcut for something, I subcontiosly press `?`. 
I think, that for easy start, it will be necessary to have this functional, fore users to be able to show help really easily from keyboard, and also close easy from keyboard (Esc). 

I also leave this in 1.0 milestone, because I think, it should be there when we show it to colegues.

jakuub:
Autogeneration from spec is ideal scenario. For the launch, I think we don't have to autogenerate them from spec, they can be "hardcoded" into the codebase, and we can leave autogeneration for next task.

I created task for that: #103

jakuub:
Autogeneration is less important then the rest of the 1.0 issues :-)

PR #37, opened 2018-11-05
Implement help button (show spec)

Siegrift:
Show a spec page on a new tab using cmd+h or clicking a help icon.

#33


src/common/Header.js
```diff
@@ -53,6 +54,15 @@ const Header = ({ user, classes }) =>
           </Avatar>
         )}
 
+        <Button
+          className={classes.item}
+          variant="outlined"
```
mikx98: maybe this: https://material-ui.com/demos/buttons/#floating-action-buttons would be nicer :thinking:

src/common/Header.js
```diff
@@ -53,6 +54,15 @@ const Header = ({ user, classes }) =>
...
+          color="inherit"
+          href={SPEC_URL}
+          target="_blank"
```
mikx98: :pray: https://medium.com/@ali.dev/how-to-fix-target-blank-a-security-and-performance-issue-in-web-pages-2118eba1ce2f

src/worklogs/WorkLogs.js
```diff
@@ -82,6 +82,9 @@ export default compose(
       clearSelection,
       setJiraIssueForSelection
     }) => e => {
+      e.preventDefault()
+      e.stopPropagation()
```
mikx98: nope, we don't want this - you won't need it after rebase to master

src/worklogs/WorkLogs.js
```diff
@@ -153,6 +156,11 @@ export default compose(
           }
           break
 
+        case 'h':
+          if (e.ctrlKey) {
+            window.open(SPEC_URL, '_blank')
```
mikx98: the security issue :top:

mikx98:
Also, please mark that this PR closes the related issue

src/common/Header.js
Siegrift: I used the IconButton, I think it looks better now (FABs were not looking good)

src/common/Header.js
Siegrift: :+1:

Siegrift:
I've changed the shortcut, because we are fighting with default browser actions (I've tested on chrome and firefox developer)

Issue #32, opened 2018-11-02
Undo/Redo


mikx98:
Implement undo/redo functionality for actions changing `store.points` (probably just that one)

PR #40, opened 2018-11-05
Undo redo

Siegrift:
#32


src/store/state.js
```diff
@@ -27,7 +29,9 @@ const getInitialState = (): State => ({
   cursor: 0,
   selection: [],
   dialog: null,
-  issues: []
+  issues: [],
+  history: [],
+  historyIndex: -1
```
mikx98: please make just one state variable out of it. for example `history: {states: [], index: -1}`

src/store/actions.js
```diff
@@ -176,14 +189,67 @@ export const editTimeForSelection = makeSelectionAction(editTime)
...
   dispatch(updateValue(['issues'], issues))
 }
+
+export const addToHistory = (filteredState: Object) => ({
+  type: 'Add to history',
+  payload: filteredState,
+  reducer: (state: State) => {
+    const { history, historyIndex } = state
+    const newHistory = [...history]
+    newHistory.splice(historyIndex + 1, history.length, filteredState)
```
mikx98: do not mutate const variable, chain the splice function to the line above

src/store/actions.js
```diff
@@ -176,14 +189,67 @@ export const editTimeForSelection = makeSelectionAction(editTime)
...
+})
+
+export const undo = () => ({
+  type: 'Undo',
+  reducer: (state: State) => {
+    if (state.historyIndex <= 0) {
+      return state
+    }
+    const { history, historyIndex } = state
+    const filteredState = history[historyIndex - 1]
```
mikx98: why filtered? let's not bring the details that it's not the whole state here, rather say what it is .. it is simple previous state (maybe just part of it, but does it matter?)

src/store/actions.js
```diff
@@ -176,14 +189,67 @@ export const editTimeForSelection = makeSelectionAction(editTime)
...
+})
+
+export const redo = () => ({
+  type: 'Redo',
+  reducer: (state: State) => {
+    if (state.historyIndex === state.history.length - 1) {
+      return state
+    }
+    const { history, historyIndex } = state
+    const filteredState = history[historyIndex + 1]
```
mikx98: the same :top:

spec/src/index.adoc
```diff
@@ -96,4 +96,4 @@ kbd:[d]:: Set description (work-log)
 kbd:[t]:: Set JIRA issue
 kbd:[n]:: Set JIRA issue to _None_
 kbd:[{commandkey} + z]:: Undo
-kbd:[{commandkey} + r]:: Redo
+kbd:[{commandkey} + y]:: Redo
```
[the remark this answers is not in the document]
Siegrift: I am also adding a possibility for command + shift + z

mikx98:
Also, please mark that this PR closes the related issue to automate the process in `Project -> Priorities`

src/App.js
```diff
@@ -12,7 +16,10 @@ import WorkLogs from './worklogs/WorkLogs'
 import type { User } from './store/state'
 
 type Props = {|
-  user: ?User
+  user: ?User,
+  fetchUser: () => void,
+  checkpoint: () => void,
+  fetchIssues: () => void
```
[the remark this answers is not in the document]
Siegrift: You can flow it (to some degree), but we are not doing so. But I see your point and will remove the functions from props (but also the | to mark that the props are not exact). The other option would be to convert this to a class and then we could get a better coverage, but I don't think that's necessary.

src/store/actions.js
Siegrift: You can't chain it, the splice function will return an array of removed elements, and it's splicing in-place.

src/App.js
```diff
@@ -27,17 +34,23 @@ export default compose(
...
+    async componentDidMount() {
+      const { user, fetchUser, checkpoint, fetchIssues } = this.props
 
       if (user) {
         return
       }
 
-      fetchUser()
+      await fetchUser()
+      await fetchIssues()
```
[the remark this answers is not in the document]
Siegrift: The reason was to fetch it everything when you have the user. And at this point the user is logged in right (otherwise the if above will return from the function)?

src/store/actions.js
Siegrift: Yeah, that's a good idea (no it doesn't matter that it's not a complete state).

src/App.js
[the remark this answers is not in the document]
mikx98: yop, agree :+1:

src/App.js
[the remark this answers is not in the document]
mikx98: hmm, that sounds right :) okay then

src/store/actions.js
mikx98: aaah, that's weird function :cry:

mikx98: i don't like it, maybe filtering it by index and appending new value?

src/worklogs/WorkLogs.js
```diff
@@ -31,6 +32,7 @@ const getMetaKeyPath = (
   metaKey: boolean,
   shiftKey: boolean
 ) => {
+  if (ctrlKey && shiftKey) return 'withCtrlShiftKey'
```
mikx98: `(ctrlKey || metaKey) && shiftKey`

src/worklogs/WorkLogs.js
```diff
@@ -129,7 +138,8 @@ export default compose(
   withHandlers({
     handleKeyDown: ({ mode, keyBindings }) => e => {
       const metaKeyPath = getMetaKeyPath(e.ctrlKey, e.metaKey, e.shiftKey)
-      const path = [mode, metaKeyPath, e.key]
+      console.log(metaKeyPath, e.key, e.key.toLowerCase())
```
mikx98: nope

src/store/actions.js
```diff
@@ -176,14 +201,72 @@ export const editTimeForSelection = makeSelectionAction(editTime)
...
 }
 
-export const fetchIssues = () => async dispatch => {
+export const fetchIssues = () => async (dispatch: Dispatch) => {
   const response = await fetch('/api/issues')
   const issues = response.status < 400 ? await response.json() : []
   dispatch(updateValue(['issues'], issues))
 }
+
+export const addToHistory = (filteredState: Object) => ({
```
mikx98: `filteredState` again :cry:

mikx98: `stateToCapture` or something like that :pray:

PR #42, opened 2018-11-05
Add flow check script

Siegrift:
I wrote a simple script a few months ago, it scans directories recursively for "// @flow" and reports files that are not flowed. I find it a good way to remember flowing the files.

Unfortunately, I couldn't set it up as a precommit hook, bacause standart output is captured by the hook service.


check_flow.py
```diff
@@ -0,0 +1,24 @@
+from os import walk, path
+from glob import glob
+
+README = '''
+Recursively searching for *.js files inside specified folders and 
+checking if the file is flowed ('// @flow' is on the first line)
+'''
+DELIM = '---------------------------------------------------------------'
+PATHS = ['src/common', 'src/login', 'src/store', 'src/types', 'src/worklogs']
```
mikx98: why not all js files under `src/`?

check_flow.py
```diff
@@ -0,0 +1,24 @@
...
+
+print(README)
+print(DELIM)
+
+files = []
+for PATH in PATHS:
+    files.extend([y for x in walk(PATH) for y in glob(path.join(x[0], '*.js'))])
```
mikx98: better naming? directory for x and file or filename for y?

check_flow.py
```diff
@@ -0,0 +1,35 @@
+from os import walk, path
+from glob import glob
+
+README = '''
+Recursively searching for *.js files inside specified folders and
+checking if the file is flowed ('// @flow' is on the first line)
+'''
```
mikx98: I would put `// no flow` possibility also here in README

src/Root.js
```diff
@@ -1,13 +1,30 @@
...
+type RootProps = {
+  store: Store
+}
+
+type DispatchProviderProps = {
+  dispatch: Dispatch,
+  children: React.Node
+}
+
+type ErrorBoundaryState = {
```
mikx98: exact type pls

src/Root.js
```diff
@@ -21,21 +38,23 @@ class DispatchProvider extends React.Component {
...
-  // eslint-disable-next-line handle-callback-err
-  componentDidCatch = (error, info) => {
-    this.setState({ hasError: true, error, info })
+  componentDidCatch(error, info) {
+    this.setState({ error, info })
   }
 
   render() {
-    if (this.state.hasError) {
+    if (this.state.error && this.state.info) {
```
mikx98: :+1:

src/index.js
```diff
@@ -1,27 +1,26 @@
...
 
-const rootElement = document.getElementById('root')
+const rootElement = (document.getElementById('root'): any)
 
 ReactDOM.render(<Root store={store} />, rootElement)
 
 // a few modules already required by `main.js` won't be updated
+// $FlowFixMe module.hot will exist if hot reload plugin is enabled
 if (module.hot) {
+  // $FlowFixMe module.hot always contains accept method
```
mikx98: :joy:

src/store/history.js
```diff
@@ -1,8 +1,13 @@
...
 import { addToHistory } from './actions'
 import { filterObject } from './utils'
 import { UNDOABLE_STATE } from '../consts'
 
-const historyMiddleware = ({ dispatch, getState }) => next => action => {
+import type { Dispatch, MiddlewareProps, GenericAction } from './reduxTypes'
+
+const historyMiddleware = ({ dispatch, getState }: MiddlewareProps) => (
+  next: Dispatch
+) => (action: GenericAction<*, *>) => {
```
mikx98: don't use `*`, use `any`. There is some subtle difference, sometimes resulting in many many errors that you don't know where it came from.

src/store/reduxTypes.js
```diff
@@ -5,11 +5,11 @@ export type PathElement = number | string
...
-  +type: string,
-  +path?: Path,
-  +payload?: Payload,
-  +reducer: SegmentReducer<Segment, Payload>,
-  +doNotLog?: boolean
+  type: string,
+  path?: Path,
+  payload?: Payload,
+  reducer: SegmentReducer<Segment, Payload>,
+  undoable?: boolean
```
mikx98: or function, based on that code you provided

src/Root.js
Siegrift: If I put exact types here I get an flow error on this.state = {} in component :/. 
I've added exact types to other type defintions in the fine though (and left this as is).

check_flow.py
Siegrift: :+1:

PR #48, opened 2018-11-07
Split actions, move reduxTypes.js and utils.js

Siegrift:
#43, I've also moved utils to the root dir, and reduxTypes to types folder. I didn't do a great job with splitting the actions, because all of them seemed to be worklog actions. If you see a better way how to split it pls let me know.


mikx98:
@Siegrift pls fix flow + there are unflowed files

PR #54, opened 2018-11-08
Tempo auth

### LINKED ISSUE #34: Tempo auth

mikx98:
Resolves #34


.env.template
```diff
@@ -23,5 +23,17 @@ ft_jira_request_url=https://vacuum.atlassian.net/plugins/servlet/oauth/request-t
 ft_jira_access_url=https://vacuum.atlassian.net/plugins/servlet/oauth/access-token
 ft_jira_authorize_url=https://vacuum.atlassian.net/plugins/servlet/oauth/authorize?oauth_token=
 
-ft_jira_api_call_myself=https://vacuum.atlassian.net/rest/api/3/myself?expand=groups
+# Do not use in production!
```
Siegrift: New entries are being added here. Should I get my config updated somehow? Or it's everything fine for now?

package.json
```diff
@@ -53,6 +53,7 @@
     "redux-localstorage-simple": "^2.1.4",
     "redux-logger": "^3.0.6",
     "redux-thunk": "^2.3.0",
+    "request-promise": "^4.2.2",
```
Siegrift: Why are we not using fetch instead? https://github.com/bitinn/node-fetch
Is it because of Bluebird?

README.md
```diff
@@ -14,3 +14,12 @@ Take a look at ["raw" (unconverted) version of specs (current branch)](./spec/sr
 ## Usage
 
 1.  `yarn start` - it will concurrently run FE and BE server, both in dev HMR mode
+
+## Development tips
+
+- Tempo oAuth doesn't work really well in development. You have two choices:
+
+  - You can override access token via `ft_tempo_override_access_token` env variable (**recommended**).
```
Siegrift: Override it with what?

server/apiCallbacks.js
```diff
@@ -60,3 +57,20 @@ export const getIssueList = async (req, res) => {
...
+
+export const getWorklogs = async (req, res) => {
+  const request = getTempoRequest(req.session)
+
+  const data = await request({
+    method: 'GET',
+    uri: config.tempo.api.worklogs,
+    qs: {
+      username: req.session.jira.userId,
+      dateFrom: '2018-11-08',
```
Siegrift: Is the hardcoded date here for a reason?

.env.template
mikx98: yes, you need to add these also to your `.env` file. It should reflect your `.env`, the only difference is that it lacks sensitive data

package.json
mikx98: yop, it's because of bluebird

README.md
mikx98: with the token you provide to the variable, f.e. `ft_tempo_override_access_token=iuh8nkj18yhbnand1d334`

server/apiCallbacks.js
mikx98: nope, just for testing purposes.. we don't use this function as of now, but will change to `req.params`

Siegrift: I don't really like pushing unused code :/ But we can probably live with that

PR #63, opened 2018-11-18
Issues performance

### LINKED ISSUE #60: Speed up issues operations
- investigate why it's so slow
- fetch parallelly, cache issues on server
- speed up FE search
- investigate rerendering
- slicing results (?)

reason: there are people that have access to over 11k issues, the fetching and searching logic is too slow for them

mikx98:
- BE perf:
  - parallel fetching of issues, in batches of 5 requests
  - server caching of all issues per user (app loads cached issues at first, then it polls on issues in case something changed)
  - fetching logic rewritten to data-providers

- FE perf:
  - stick with native components
  - slice results (avoid unnecessary renders - react-virtualized wasn't enough)
  - added debouncing

Resolves #60


package.json
```diff
@@ -42,7 +44,7 @@
     "passport-local": "^1.0.0",
     "pg": "^7.4.3",
     "postcss-loader": "2",
-    "react": "^16.0.0-rc.3",
```
Siegrift: LUL, we were using old release candidate xD

src/App.js
```diff
@@ -27,28 +25,30 @@ const App = ({ user }: Props) => (
...
+      getData: [fetch, 'api/issues?cache=true', { credentials: 'same-origin' }],
+      onData: [dispatchReceivedData, ['issues']],
+      needed: false
+    },
+    {
+      ref: 'issues (new)',
+      getData: [fetch, 'api/issues', { credentials: 'same-origin' }],
+      onData: [dispatchReceivedData, ['issues']],
+      polling: 20 * 60 * 1000,
+      fetchTimeout: 3 * 1000,
```
Siegrift: These are also set by the default data provider config and can be removed

mikx98: oh, that constant should have been much bigger :thinking: i wanted to explicitly specify that this fetch is allowed to take more than 30s (default). will set to 2 minutes

PR #75, opened 2018-12-18
Speed dial

### LINKED ISSUE #31: Jira issues speed dial
Probably everyone has these issues he repeats most of the time. Add some key bindings for kind of "speed dial" for them.

* they can be set either via custom key bindings (ctrl+shift+1, 2, ...) or via settings (accessible via Header, after clicking on avatar?)
* they would be used via easier key bindings (ctrl+1, ctrl+2, ...)
* update the spec

* consider implementing "default" issue - by default every worklog will be associated with that issue (set: settings or ctrl+shift+0, use: automatic)

mikx98:
Closes #31


src/worklogs/WorkLogs.js
```diff
@@ -119,7 +124,16 @@ export default compose(
             s: () => openDialog(DIALOG_TYPES.SUBMIT_WORKLOGS),
             '-': clearSelection,
             z: undo,
-            y: redo
+            y: redo,
+            ...[...Array(10).keys()]
```
Siegrift: Maybe create a custom function for that (so that you don't have to understan implementation)

mikx98: check once again pls :pray:

PR #88, opened 2019-01-04
Chrome fix

Siegrift:
Fix button positioning on chrome (shows a little differently on firefox and chrome, due to inconsistent behaviour of browsers)


src/worklogs/dialogs/SubmitWorklogsDialog.js
```diff
@@ -21,17 +22,42 @@ import Check from '@material-ui/icons/Check'
 import Loading from '../../common/Loading'
 import FinalWorklogsCalendar from '../calendar/FinalWorklogsCalendar.js'
 
+const styles = theme => ({
+  dialog: {
+    display: 'block !important'
```
mikx98: ;)

PR #95, opened 2019-01-09
Multiday logging

### LINKED ISSUE #83: Multiday logging
While still supporting just logging one day, add functionality to log to not just TODAY, but even go back in history and for example log YESTERDAY, etc.

- [x] calendar navigation
- [x] inserting to currently shown day/today

mikx98:
Closes #83


src/worklogs/calendar/calendarSelectors.js
```diff
@@ -1,13 +1,18 @@
...
 import worklogsSelector from '../../selectors/worklogsSelector'
 import { MODES } from '../../consts'
 
 import type { State } from '../../store/state'
 
+const ETERNITY = moment()
+  .add(1000, 'years')
+  .toDate()
+
 export type Event = {
```
Siegrift: I don't like the "Event" naming as that is too general and is already used be browsers.

src/worklogs/calendar/calendarSelectors.js
```diff
@@ -36,19 +62,21 @@ export const calendarEventsSelector: InputSelector<
...
-    }))
+    points
+      .map((point, i) => ({
+        issue: point.issue,
+        title: point.description,
+        index: i,
+        start: new Date(point.timestamp),
+        end:
+          i === points.length - 1
+            ? ETERNITY
```
Siegrift: Why this change? From now on the last thing we log will have it's endpoint in 1000 years?

src/worklogs/calendar/config.js
```diff
@@ -7,7 +7,7 @@ import WorkLogEvent from './WorkLogEvent'
 const calendarConfig = {
   localizer: Calendar.momentLocalizer(moment),
   defaultView: 'day',
-  toolbar: false,
+  toolbar: true,
   views: { day: true },
```
Siegrift: add agenda: true. I like it and as the toolbar is already there. I don't see disadvntage :)

Siegrift:
And also when I am on a different day (in calendar) and I log some point, I would expect it to be logged to the day I am currently in UI (and not today)

src/worklogs/calendar/calendarSelectors.js
mikx98: in order to indicate that this thing is really IN PROGRESS and does not have an end date. Visually, that also means, that it will be up in the calendar also when you navigate the calendar to the future

src/worklogs/calendar/config.js
mikx98: no :D it looks bad, so i don't want to add it there until some styling is done

mikx98:
Okay, but what time do you expect to log it at? Probably not this time yesterdays' time if you navigated to yesterday... but maybe if it will open insert point in time dialog instead of just logging, when simple logging in different days' view?

src/worklogs/calendar/calendarSelectors.js
mikx98: the react big calendar component we are using calls it event, hence the name .. i wanted to indicate that this is the actual type we supply there. what do you mean by browsers' events?

src/worklogs/calendar/config.js
Siegrift: Mhh it doesn't make it worse I believe. I will try to work on the design... The toolbar is quite fine imo, however the work in progress is really ugly and not user friendly. I will also try another libraries :)

mikx98: okay :+1: PRs are always welcome, but this is certainly not a blocker for this PR :)

jakuub:
I agree with Emo, i expect the same - I log into the day I can see. I understand, that insert is more suitable for this situation, so I suggest, to add some "warning" next to logging, that it will log to "today" and if someone want to log to currently shown date, she should use insert functionality.

jakuub:
@Siegrift I want to merge this and create another task with the small information about logging day. From code perspective, is it OK for you to merge this? 

The state - log current time to "today" and log with "insert" mode by default to "visible day" is OK for me, if I can see to which day I am loggin each time I do it.

PR #97, opened 2019-01-10
Smart jira issues search

### LINKED ISSUE #94: smarter JIRA issue search
Search should not be sensitive to a word order, i.e. `project management` and `management project` should be considered match.

This is especially important when searching for issues that contain similar words in multiple projects. In such a case, one would really like to use query like `WOR project management` to search for project managament task on wordy project.

The simple solution is to:

* split by words both needle and haystack
* the match is considered when each word of needle is a prefix of some word in a result

mikx98:
Closes #94


src/worklogs/components/IssueInput.js
```diff
@@ -78,11 +78,14 @@ export default compose(
...
-        const pattern = inputValue.toLowerCase()
+        const regexes = inputValue
+          .split(' ')
+          .map(value => new RegExp(`\\b${value}`, 'i'))
+
         return new Promise(resolve =>
           resolve(
             issues
-              .filter(issue => issue.label.toLowerCase().includes(pattern))
+              .filter(issue => regexes.every(regex => regex.test(issue.label)))
```
Siegrift: It's and improvement, but I would like fuzzy search implemented here. There should be many libraries for it: https://bashooka.com/coding/javascript-fuzzy-search-libraries/

jakuub:
@Siegrift mergnime to a fuzzy search spravme neskor.

jakuub:
I discussed this with @Siegrift , @mikx98 we can merge it :-).

PR #117, opened 2019-01-24
Reorder cheat sheet

mikx98:


src/worklogs/keyBindingsUtils.js
```diff
@@ -108,50 +106,50 @@ const getKeyBindings = ({
...
         '?': {
           handler: () => openDialog(DIALOG_TYPES.CHEAT_SHEET)
         }
       },
-      withAltKey: getSpeedDialHandlers(
-        openDialog,
-        DIALOG_TYPES.SET_SPEED_DIAL_OPTION
-      ),
+      withAltKey: {
+        '[0-9]': { description: 'Set speed dial option' },
```
Siegrift: It is only for documentation right? This is not valid keybinding

mikx98: Yes, just documentation, it's impossible to press everything together :)

PR #120, opened 2019-01-28
Update logo

hleumas:
So it's nice.


Siegrift:
I don't know if its by design, but the logo spans the whole screen (which I think is not that nice, as button is still really small)

hleumas:
@Siegrift this should fix that.

PR #137, opened 2019-02-15
Assign issues with ctrl + number

### LINKED ISSUE #136: Assign speed dial on CTRL+number instead of ALT+number
The problem of ALT is that the combination of ALT+number is sometimes (on macs) interpreted as a special character, not as the combination - therefore assigning issues to speed dial is not accessible on MacBooks. 

This could partially help, but not fully: https://github.com/vacuumlabs/fine-tracker/issues/111

I mark this as a bug because speed dial assignment does not work on Macs.

mikx98:
Closes #136


jakuub:
Now I realized something - this will not work on my computer - CTRL+1 switches me to the first tab. 

I don't know what is the best solution, how to fix this for both - apple users, and users like me. Maybe both shortcuts, or combination with shift (or tab like asana :-D). Maybe @darjanin will have a good idea.

mikx98:
try it in localhost, but i think that it should consume the event - so browser shortcut will not get caught :). Unless you mean that you are actually using the browser shortcut :smile:

jakuub:
Ok, thank you :-)

PR #139, opened 2019-03-09
Information about logging day

### LINKED ISSUE #104: Information about logging day
Information next to the logging input about the fact, that it will log to "today", not "currently opened day". 

Originaly from https://github.com/vacuumlabs/fine-tracker/pull/95#issuecomment-453967578

mikx98:
Closes #104


jakuub:
Do we really want to do it right now, when the new design is on the way? I don't know if it interferes somehow, but it looks like this will be discarded in the first attempt of the new design.

mikx98:
No idea, IMO if it gets removed, it will get removed :D it's not such a big thing

PR #143, opened 2019-04-09
Refactored code structure

domiin50:


src/App.js
```diff
@@ -26,15 +26,15 @@ import {
...
 import FocusLock from 'react-focus-lock'
-import Header from './common/Header'
-import Login from './login/Login'
-import WorkLogInput from './worklogs/components/WorkLogInput'
-import WorkLogCalendar from './worklogs/calendar/WorkLogCalendar'
-import DialogContainer from './worklogs/dialogs/DialogContainer'
+import Header from './Tracker/Header'
+import Login from './Login'
+import WorkLogInput from './Tracker/Logger/LoggerInput'
+import WorkLogCalendar from './Tracker/Timeline/WorkLogCalendar'
```
mikx98: each directory except for the files / and or directories containing a root component should start with a lower case letter - components are the only ones allowed to start with an upper case one :) so if you are planning to have a root component (`index.js`) for Tracker, Logger, Timeline, ... everything's okay, if no, lower the case of the corresponding directory :)

PR #146, opened 2019-05-20
Added content of Shortcuts component

domiin50:


src/tracker/Shortcuts/index.js
```diff
@@ -11,28 +11,88 @@ import './shortcuts.css'
 
 type Props = {|
   shortcutsShown: boolean,
-  toggleShortcuts: (shortcuts: boolean) => void
+  toggleShortcuts: (shortcutsShown: boolean) => void
 |}
 
+const SHORTCUTS = {
```
mikx98: Try to have all the shortcuts in one place only. Currently there is one more place - the handler. Maybe adding a category there is all that we need :)

src/tracker/Shortcuts/index.js
```diff
@@ -1,38 +1,92 @@
...
-      <div onClick={toggleShortcuts} style={{ paddingRight: 15 }}>
-        {shortcutsShown ? <Close /> : <CallMade />}
-      </div>
+const mapPathToSpecialKeys = (path: string): string[] =>
+  // $FlowFixMe
+  [
+    path.toLowerCase().includes('ctrl') && 'CTRL',
+    path.toLowerCase().includes('alt') && 'ALT',
+    path.toLowerCase().includes('shift') && 'SHIFT'
+  ].filter(Boolean)
```
mikx98: even shorter :):
```suggestion
  ['ctrl', 'alt', 'shift'].map(key => path.toLowerCase().includes(key) && key.toUpperCase()).filter(Boolean)
```

src/tracker/Shortcuts/index.js
```diff
@@ -1,38 +1,92 @@
...
+
+const getShortcuts = (keyBindings: Object) =>
+  _(
+    Object.values(MODES).map((mode: any) =>
```
mikx98: this first map can be simplified by iterating through the values of keyBindings - notice that everywhere you use mode, it's always in keyBindings[mode] context. So mapping it by values would simplify that.
`Object.values(keyBindings).map(bindings => Object.keys(bindings).map(path => {...))` so each `keyBindings[mode]` would be replaced by `bindings`

PR #148, opened 2019-05-23
Refactoring and redesign of WorkLogs component

domiin50:
A little refactoring of some props according to WorkLogs component. Also updated and refactored design of component into a CSS file.
Fixed promise rejects value when the provider of issues crashes with an error.


src/tracker/WorkLogs/index.js
```diff
@@ -16,60 +16,41 @@ import type { Event } from '../Timeline/timelineSelectors'
...
-        overflowY: 'auto',
-        height: `calc(100vh - ${shortcutsShown ? '590px' : '410px'}`
-      }}
+      className="container"
+      style={{ height: `calc(100vh - ${shortcutsShown ? '590px' : '395px'}` }}
     >
       {events.map((event, index) => (
         <div
-          className="worklog"
+          className={`worklog ${event.selected ? 'selected' : ''}`}
```
mikx98: use `classnames` library for that

src/tracker/WorkLogs/index.js
```diff
@@ -79,10 +60,10 @@ const WorkLogs = ({
 export default compose(
   connect(
     state => ({
-      points: state.points,
-      dateShown: state.dateShown,
       events: eventsSelector(state),
-      shortcutsShown: state.shortcutsShown
+      displayedDate: state.displayedDate,
+      shortcutsShown: state.shortcutsShown,
+      _state: state
```
mikx98: this is not what you want... do you actually use it? or is it just a legacy from debugging?

domiin50: legacy.. x(

PR #151, opened 2019-05-27
Redesigned and refactored Timeline component

domiin50:
Redesign and refactor of Timeline component. Also, need to talk about selectors splitted in refactor PR. It is needed to select first all of the events and then filter them not in reverse, because of the indexes of events in Timeline and WorkLogs overview component.


domiin50:
The reason for the failed test: Forgot to clear selector that it is needed to talk about. It is assigned but not used right now.

src/tracker/Timeline/index.js
```diff
@@ -13,145 +13,86 @@ import {
...
+    </div>
+    <div className="intervals">
+      {events.map((event, index) => {
+        const eventEnd = event.actual ? moment() : event.end
+        const spacing = countSpacing(
+          moment(eventEnd).diff(moment(event.start), 'minutes')
+        )
+        return (
+          <div
+            onClick={() => toggleSelectPoint(event.index)}
```
mikx98: memoize bind

src/tracker/Timeline/index.js
```diff
@@ -13,145 +13,86 @@ import {
...
+        .map((_, index) => (
+          <div className="point" key={index}>
+            <small className="label">
+              {`${(index + startHour) % TIME_FORMAT}`.padStart(2, '0')}:00
+            </small>
+          </div>
+        ))}
+    </div>
+    <div className="intervals">
+      {events.map((event, index) => {
```
mikx98: seems to me that this map has too much logic inside.. try to make a separate small component for that and map directly to the component here

src/tracker/Timeline/timelineEventsSelector.js
```diff
@@ -20,4 +20,27 @@ const filteredPointsSelector: InputSelector<
     points.filter(point => moment(point.timestamp).isSame(displayedDate, 'day'))
 )
 
-export default getEventsSelector(filteredPointsSelector)
+// export default getEventsSelector(filteredPointsSelector)
```
mikx98: please keep removing the commented code :) i will be stricter and stricter about that :smile:

src/tracker/Timeline/timelineEventsSelector.js
```diff
@@ -20,4 +20,27 @@ const filteredPointsSelector: InputSelector<
...
+    const filteredEvents = events.filter(event =>
+      moment(event.start).isSame(displayedDate, 'day')
+    )
+    if (!filteredEvents.length)
+      return [
+        Object.assign({}, events[events.length - 1], {
+          start: moment(displayedDate).startOf('day')
+        })
+      ]
+    return filteredEvents
```
mikx98:
```
return filteredEvents.length
   ? filteredEvents
   : [{...events[events.length - 1], start: moment(displayedDate).startOf('day')}]
```

PR #153, opened 2019-06-03
Added modes switching and work logs editing functionality

domiin50:


src/App.js
```diff
@@ -67,16 +67,21 @@ const App = ({ mode, dialog, isAnotherDay, createPoint }: Props) => (
...
+      <div onClick={handleClick}>
+        <div
+          className={classNames({ 'dark-overlay': mode === MODES.LOGGING })}
+        />
+        <Timeline />
+        <div style={{ display: 'flex', height: 'calc(100vh - 265px)' }}>
+          <div style={{ flex: 1 }}>
+            <WorkLogs />
+            <Shortcuts />
+            {/*<div style={{ marginBottom: '56px' }}>
```
mikx98: Remove it, we will never use it again :)

src/tracker/Logger/index.js
```diff
@@ -61,25 +78,58 @@ const Logger = ({
...
+      selection,
       createPoint,
+      editDescription,
+      editTime,
+      clearSelection,
       setInputOverride
     }) => () => {
+      if (mode !== MODES.LOGGING) {
+        editDescription(description)
+        if (selection.length === 1) editTime(time)
```
mikx98: So we stopped supporting the change of time of multiple?

src/tracker/utils.js
```diff
@@ -2,14 +2,15 @@
 import moment from 'moment'
 
 export type Time = { hours: number, minutes: number }
+export type Override = { description: string, time: Date }
```
mikx98: PropsOverride, Override is way too generic for a name

src/tracker/Logger/index.js
mikx98: Also, what about trying to change the time of multiple - will I reset the description of each one to the description of the first one?

src/App.js
domiin50: It is done in one of the next commits :)

src/tracker/Logger/index.js
domiin50: I had a meeting with Milan and we agree together that it should not be able to change the time of multiple points to be the same. It does not make any sense.

domiin50: Probably it would be better to disable the time input when selected multiple work logs. This will prevent the case you are talking about.

PR #159, opened 2019-06-06
Added cursor functionality

domiin50:


src/tracker/Logger/index.js
```diff
@@ -113,7 +114,7 @@ export default compose(
     }) => () => {
       if (mode !== MODES.LOGGING) {
         editDescription(description)
-        if (selection.length < 2) editTime(time)
+        firstSelectedPoint && selection.length < 2 && editTime(time)
```
domiin50: In the next PR, there will be one fix. FrstSelectedPoint = first from selection or cursor. It can be null in a specific case.

PR #162, opened 2019-06-06
Refactored keys handling in inputs

domiin50:


src/tracker/Logger/LoggerTimeInput.js
```diff
@@ -78,6 +80,7 @@ const LoggerTimeInput = ({
       onFocus={handleFocus}
       onKeyDown={handleKeyDown}
       onValueChange={handleChange}
+      onChange={() => setModified(mode === MODES.LOGGING)}
```
mikx98: memoize-bind

src/App.js
```diff
@@ -125,9 +125,17 @@ export default compose(
...
         e.altKey
       )
-      const path = [mode, metaKeyPath, e.key.toLowerCase(), 'handler']
+      const path = [mode, metaKeyPath, e.key.toLowerCase()]
+      const { handler, inputFriendly } = get(keyBindings, path) || {}
 
-      const handler = get(keyBindings, path)
+      if (
+        e.target.tagName === 'INPUT' &&
+        !INPUT_FRIENDLY_MODES.includes(mode) &&
```
mikx98: the keyBinding already contains information about the mode where it is enabled, do not introduce this constant

src/App.js
```diff
@@ -125,9 +125,17 @@ export default compose(
...
+        !inputFriendly
+      ) {
```
mikx98: imho this whole condition is really hardly understandable, please write it more intuitively - what exactly means that some key binding is inputFriendly? we are stopping propagation here - how does this correlate with **input friendliness**? try to name it according to the action it is used for. f.e.: allow to execute from input

src/App.js
```diff
@@ -125,9 +125,17 @@ export default compose(
...
+        e.stopPropagation()
```
mikx98: if i understand correctly, there is no handler for key bindings on a higher level than this one. use `e.preventDefault()` to prevent browser doing its custom action. stop propagation shouldn't do anything imo

PR #169, opened 2019-06-24
Implement getting JIRA issues through SSO proxy.

domiin50:
Reimplement JIRA issue API to be demanded to VLSSO and not right to JIRA
API. To be handled as proxy request by VLSSO.


server/apiCallbacks.js
```diff
@@ -21,39 +20,37 @@ export const getLoggedUserInfo = (req, res) => {
...
-        'GET',
-        `${
-          config.jira.api.issues
-        }&maxResults=${ISSUE_BATCH_SIZE}&startAt=${startAt}`,
-        null,
-        '',
-        'application/json',
-        getHandler(resolve, reject)
-      )
+      request({
```
mikx98: please use async - await syntax and try - catch it for exception

server/apiCallbacks.js
```diff
@@ -21,39 +20,37 @@ export const getLoggedUserInfo = (req, res) => {
...
-  const getHandler = (resolve, reject) => (error, data) => {
-    if (error) {
-      // reject(JSON.stringify(error))
-      console.error(JSON.stringify(error))
-      reject([])
-      return
-    }
-
-    const dataObj = JSON.parse(data)
+  if (!req.cookies.ssoToken) {
```
mikx98: I would like this check to be extracted into a middleware. See `checkIfLoggedIn` for more information

PR #173, opened 2019-06-26
Fix editing time of unsynced worklog after syncing

domiin50:
Issue: After syncing some work logs, it should not let to change the time of
an unsynced work log that could interfere with already synced intervals.

Handle this issue by disabling time input for problematic relevant work logs.
Also handling time editing of other work logs to not interfere with
already synced work logs.


src/actions/trackerActions.js
```diff
@@ -60,22 +57,28 @@ const editDescription = (indices: number[], description: string) => ({
...
+const handleEditingTime = (points: Point[], payload) => {
+  const newPoints = sortBy(
+    points.map((point, i) =>
+      payload.indices.includes(i)
+        ? { ...point, timestamp: payload.timestamp }
+        : point
+    ),
+    ['timestamp']
+  )
+  const myIndex = newPoints.findIndex(p => p.timestamp === payload.timestamp)
```
mikx98: Actually I really like this line :)

src/tracker/Logger/firstSelectedPointSelector.js
```diff
@@ -1,18 +1,22 @@
...
 import { createSelector } from 'reselect'
+import { touchesSyncedInterval } from '../../utils'
 import { MODES } from '../../consts'
 import type { State } from '../../store/state'
 
 const modeSelector = state => state.mode
 const pointsSelector = (state: State) => state.points
-const selectionSelector = (state: State) => state.selection
-const cursorSelector = (state: State) => state.cursor.index
+const indexSelector = (state: State) =>
```
mikx98: Maybe editIndex would be a better name?

Issue #167, opened 2019-06-18
Differentiate between submitted and yet to submit worklogs


mikx98:
It would be nice to know which worklogs are already submitted and which ones are not.

Issue #175, opened 2019-06-27
I can select (and modify / delete) oldest submitted log


MilanJimi:
1. Create a log and submit it
2. Select the oldest log on timeline
3. This log is then for example deletable with `Delete` button. 

Deleting any other log later will lead to error `TypeError: "e[(t - 1)] is undefined"`

PR #185, opened 2019-06-28
Update worklogs styling to distinguish synced ones

domiin50:
Resolves issue #167 and #175.


domiin50:
> @domiin50 How did you resolve the #175? What was the cause? I don't see it in the code.
> The #167 is ok.

Synced work logs are unable to select on the timeline and delete or edit them.

domiin50:
It also resolves #187. It is now unable to select log on the timeline as in work logs overview.
It should be enough to prevent a manipulation with yet synced points. Or is it necessary to handle it also in actions @mikx98? I do not see any more ways/leaks about how the actions could be triggered.

Issue #180, opened 2019-06-27
I can create log with no description


MilanJimi:
1. Create a log with just time
2. It still appears in FineTracker. This might not yet be a problem
3. Assign an issue and try to save the log. 

Since Jira requires description for each log, the save will fail for this log

mikx98:
Empty log is actually a very handy feature - to mark end of worklog and start of a worklog that you don't want to submit. For example you don't want to write `lunch` or `end of work`. The timeline even doesn't show such worklogs - I :heart: it personally. 

**So the only issue I see in this, is that you should not be able to assign an issue to empty-description worklog.**

domiin50:
Would like to keep this issue open till the sidebar with issue input will be finally properly implemented. Issue input will be disabled when the worklog has no description. It is pointless to handle it in action imo.

Meanwhile, I will distinguish the empty-description worklogs and handle editing description to be empty when it has assigned issue yet.

mikx98:
However, bear in mind that preventing of assigning issues from the sidebar may not be enough, since you can assign issue via key shortcuts only.

domiin50:
> However, bear in mind that preventing of assigning issues from the sidebar may not be enough since you can assign issue via key shortcuts only.

Okay, I amended the commit in #190 and added handling of issue assigning to no-description worklogs.

PR #190, opened 2019-07-01
Handle assigning issues to no-description worklogs

domiin50:
Resolves issue #180.

Handle assigning issue to only worklogs with description.
Handle editing description to be not empty when it has assigned issue yet.
Distinguish the issue info of empty-description worklogs.


src/actions/trackerActions.js
```diff
@@ -13,6 +13,12 @@ import type { Dispatch, GetState } from '../types/reduxTypes'
 
 import { PREHISTORIC_DATE, MODES } from '../consts'
 
+const hasIssue = (points: Point[]) => (index: number) =>
+  index > 0 && points[index].issue
```
mikx98: so if I understand correctly, the points[0] worklog (the one in progress) can have empty description and assigned issue? why is there the `index>0` condition?

mikx98: ok, so it is not about the one in progress, but the first one - but still I don't understand why it's there

domiin50: You are right. It is a bug.

PR #193, opened 2019-07-24
Add Google Analytics event tracking

domiin50:
Track activity of an user when logging and submitting worklogs to JIRA.


src/actions/dataProviderActions.js
```diff
@@ -17,3 +18,21 @@ export const dispatchReceivedData = (
       type: `Received data: ${ref}`
     })
   )
+
+export const dispatchReceivedUserData = () => (
+  ref: string,
```
mikx98: I would not create a whole new function but rather create a middleware for the upper one. The usage would be like `onData: [saveTrackingData(dispatchReceivedData), ['user']]`

PR #194, opened 2019-07-24
Update using VLSSO for Tempo requests

domiin50:
After Tempo integration into VLSSO, it is now able to use VLSSO as a proxy
also for Tempo requests.

Delete Tempo authentication.

Update API function for submitting work logs to make a request on VLSSO
not right to Tempo server.

Refactor and clean the mess in server folder up.


server/utils.js
```diff
@@ -24,7 +27,7 @@ export function errorHandler(err, req, res, next) {
   if (err instanceof BadRequestError) {
     res.status(400).json({ error: err.message })
   } else {
-    console.error('Error', err) // eslint-disable-line no-console
+    //console.error('Error', err) // eslint-disable-line no-console
```
mikx98: Why?

PR #208, opened 2019-10-12
Set speed dial if key is not binded to an issue

### LINKED ISSUE #111: Open assing issue to speed dial if used not set speed dial
I think, it would be great, if there is no issue on used speed dial to offer to assign one.

FrantisekMichalSebestyen:
resolves #111


src/actions/trackerActions.js
```diff
@@ -284,7 +285,7 @@ export const submitWorklogs = () => async (
     `Something went wrong. Logged ${syncedIds.length} / ${
       worklogs.length
     } logs. Please try again.`
-
+
```
domiin50: This would be probably fixed in the `prettier fix` commit, but please remove the trailing spaces.
I recommend checking the changed code every time when creating PR if there are not some unwanted changes.

domiin50:
@mikx98 is right. It will need to be redone after the Sidebar implementation. So I think it would be more sensible to wait with the merge. There will be some more changes.

PR #210, opened 2019-10-15
Validate issues

### LINKED ISSUE #131: Validate issues against actually fetched issues
If some issue gets deleted and we have it in speed dial, or even some log, sync would return error .. but maybe more convenient would be to just unassign the issue from the not submitted log or remove the speed dial option.

FrantisekMichalSebestyen:
If issue is no longer up to date with JIRA, unassign it from speed dial and not synced logs.

resolves #131


src/actions/trackerActions.js
```diff
@@ -306,6 +306,32 @@ const makeSelectionAction = actionCreator => (...args: any[]) => (
   dispatch(clearSelection())
 }
 
+export const validateIssues = (issues: Object) => (
+  dispatch: Dispatch,
+  getState: GetState
+) => {
+  const state = getState()
+  const points = state.points
+  const speedDial = state.speedDial
```
domiin50:
```suggestion
  const { points, speedDial } = getState()
```

src/actions/trackerActions.js
```diff
@@ -306,6 +306,32 @@ const makeSelectionAction = actionCreator => (...args: any[]) => (
   dispatch(clearSelection())
 }
 
+export const validateIssues = (issues: Object) => (
```
domiin50:
```suggestion
export const validateIssues = (issues: Issue[]) => (
```

src/actions/trackerActions.js
```diff
@@ -306,6 +306,32 @@ const makeSelectionAction = actionCreator => (...args: any[]) => (
...
+      ? point
+      : {
+          ...point,
+          issue: null
+        }
+  )
+
+  const validatedSpeedDial = pickBy(speedDial, key => {
+    return issueKeys.includes(key)
+  })
```
domiin50: The pickBy function could be written like `key => issueKeys.includes(key)`

PR #214, opened 2019-10-17
Refactor selection handling in inputs

domiin50:
When first selected point changes, the inputs listen for this change and
update their input values.

Refactor this handling and add SelectionListener HOC that handle changes
of the first selected point.


domiin50:
@mikx98, probably a little difficult PR for a review for @FrantisekMichalSebestyen, so please take a look.

src/tracker/firstSelectedPointSelector.js
```diff
@@ -1,8 +1,8 @@
 // @flow
 import { createSelector } from 'reselect'
-import { touchesSyncedInterval } from '../../utils'
-import { MODES } from '../../consts'
-import type { State } from '../../store/state'
+import { touchesSyncedInterval } from '../utils'
+import { MODES } from '../consts'
+import type { State } from '../store/state'
 
 const modeSelector = state => state.mode
```
FrantisekMichalSebestyen: is it okay that

domiin50:
> I like the way this is going, but I would abstract the HOC a bit more:
> let's not only listen to when the selection changes, but let's also handle the whole overriding logic in the HOC. what do you think? in short, the HOC should have it's `handleOverride` method (no other occurence of this method outside the HOC) + this HOC would be initialized by the actual setter functions (setDescription, setTime, setIssue). The responsibility of HOC would be to:
> _listen for changes of selection + override supplied inputs if necessary_

@mikx98, I do not think it is a good idea. In the issue input, the logic in `handleOverride` is a little more complicated. This function could have more logic inside than just simple setters.

.eslintrc
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": ["vacuumlabs", "prettier", "prettier/react"],
+  "extends": ["vacuumlabs", "prettier", "prettier/react", "plugin:flowtype/recommended"],
```
mikx98: hmm, what does this do?

src/tracker/withSelectionListener.js
```diff
@@ -0,0 +1,60 @@
...
+    }),
+    withHandlers({
+      resetValues: ({ handleOverride }) => () =>
+        handleOverride(getInitialInputs()),
+      setValues: ({
+        firstSelectedPoint,
+        handleOverride,
+        setIsTimeModified
+      }) => () => {
+        if (!handleOverride) return
```
mikx98: you can probably remove this line

.eslintrc
domiin50: It is needed for eslint acceptation of flow utilities (`$Keys` in one of the changed files file).

PR #219, opened 2019-10-22
Remove unused dependences

domiin50:


package.json
```diff
@@ -25,28 +25,20 @@
     "css-loader": "^3.2.0",
     "data-provider": "^0.3.1",
     "date-fns": "^1.30.1",
-    "dotenv": "^6.0.0",
```
mikx98: are you sure with this one? don't we need this for loading the variables from the .env file?

domiin50: We use our `transenv` instead. I used `depcheck` for checking which dependencies are unused and remove only those.

PR #224, opened 2019-10-23
Validate issues

### LINKED ISSUE #131: Validate issues against actually fetched issues

FrantisekMichalSebestyen:
If issue is no longer up to date with JIRA, unassign it from speed dial and not synced logs.

resolves #131
same as #210


src/actions/trackerActions.js
```diff
@@ -306,6 +306,28 @@ const makeSelectionAction = actionCreator => (...args: any[]) => (
...
 
+export const validateIssues = (issues: Issue[]) => (
+  dispatch: Dispatch,
+  getState: GetState
+) => {
+  const { points, speedDial } = getState()
+  const issueKeys = issues.map(issue => issue.key)
+
+  const validatedPoints = points.map(point =>
+    issueKeys.includes(point.issue) || point.synced
```
mikx98: this is O(n^2), i think we can do better :) there is already some selector that maps issues to some object, use that one

src/App.js
```diff
@@ -118,9 +120,11 @@ export default compose(
       setJiraIssue: _setJiraIssueForSelection,
       setJiraIssueFromSpeedDial: _setJiraIssueFromSpeedDial,
       undo: _undo,
-      redo: _redo
+      redo: _redo,
+      validateIssues: _validateIssues
     }
   ),
+
```
mikx98: remove please

src/App.js
```diff
@@ -99,6 +100,7 @@ export default compose(
     state => ({
       user: state.user,
       mode: state.mode,
+      issues: state.issues,
```
mikx98: the action is a thunk action - it can get issues by itself, no need to cause rerender of the whole app each time the issues list changes

src/utils.js
```diff
@@ -43,3 +43,6 @@ export const blurElement = () => {
 
 export const touchesSyncedInterval = (points: Point[]) => (index: number) =>
   index > 0 && points[index - 1].synced
+
+export const applyToMultiple = (fns: Function[]) => (...args: any) =>
+  fns.map(fn => fn(...args))
```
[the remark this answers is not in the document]
mikx98: what exactly doesn't work? you mean the type or what? imho no need to return anything @FrantisekMichalSebestyen - feel free to use `fns.forEach()` here, should work imo

src/App.js
```diff
@@ -83,13 +86,19 @@ export default compose(
     {
       ref: 'issues (cached)',
       getData: [fetch, 'api/issues?cache=true', { credentials: 'same-origin' }],
-      onData: [dispatchReceivedData, ['issues']],
+      onData: applyToMultiple([
+        dispatchReceivedData(['issues']),
+        postProcess(validateIssues)
+      ]),
```
domiin50: Is this needed? There should be no changes when loading from the cache, shouldn't be?

mikx98: Good point, correct, let's remove it from here :)

domiin50: @FrantisekMichalSebestyen please, take a look

PR #227, opened 2019-10-25
Fix speed dial sidebar

### LINKED ISSUE #226: New desing - fix speed dial
With a new sidebar, assigning speed dial is broken.

FrantisekMichalSebestyen:
Make speed dial work for Issue input on JIRA Sidebar.
Get rid of speed dial dialog.

resolves #226


src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -69,18 +71,30 @@ export default compose(
...
-      setJiraIssue(option ? option.value : null)
-      option && saveRecentIssue(option.value)
+    handleSubmit: ({
+      option,
+      setJiraIssue,
+      saveRecentIssue,
+      focus,
+      setSpeedDialOption
+    }) => () => {
+      if (focus.payload) {
```
domiin50: Payload of the focus is not set in the state (see comments below). Use just local prop.

Also, it is needed to reset this value on the blur of the input if there is some.
Because it is able to get to the input also just with the mouse click. So after blurring the input, the prop is still filled with the payload. And after follow focus with the mouse click we want to set the issue to worklogs again.
The input should have `onBlur` event. See `LoggerTimeInput` file.

src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -69,18 +71,30 @@ export default compose(
...
+        setSpeedDialOption(focus.payload, option ? option.value : null)
+      } else {
+        setJiraIssue(option ? option.value : null)
+        option && saveRecentIssue(option.value)
```
domiin50: Extract this line outside the condition. The option should be save to the recent issues every time it is used.
So, the whole condition could be simplified: `CONDITION && SetSpeedDialOption || SetJiraIssue`.

src/tracker/JiraSidebar/speedDialSelector.js
```diff
@@ -7,6 +7,7 @@ import type { State, Issue } from '../../store/state'
 
 export type SpeedDial = Issue & { number: string }
 
+export const focusSelector = (state: State) => state.focus
```
domiin50: This won't be needed anymore.

src/consts.js
```diff
@@ -1,12 +1,10 @@
 // @flow
 export type Mode = 0 | 1
 
-export type Focus = 'description' | 'time' | 'issue'
+export type FocusType = 'description' | 'time' | 'issue'
+export type Focus = { type: FocusType, payload?: any }
```
domiin50: Do we use this type somewhere? Is it needed?

src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -69,22 +75,57 @@ export default compose(
...
+      option,
+      setJiraIssue,
+      saveRecentIssue,
+      setSpeedDialOption,
+      payload
+    }) => () => {
+      payload
+        ? setSpeedDialOption(payload, option ? option.value : null)
+        : setJiraIssue(option ? option.value : null)
       option && saveRecentIssue(option.value)
```
domiin50: This could be refactored somehow like: `const optionValue = option ? option.value : null` and use it in all of 3 functions.

src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -69,22 +75,57 @@ export default compose(
...
     {
       setJiraIssue: _setJiraIssue,
-      saveRecentIssue: _saveRecentIssue
+      saveRecentIssue: _saveRecentIssue,
+      setFocus: _setFocus,
+      setSpeedDialOption: _setSpeedDialOption
     }
   ),
   withState('option', 'setOption', null),
+  withState('payload', 'setPayload', null),
```
domiin50: move this to the `withFocusListener` HOC

FrantisekMichalSebestyen:
Closing, because of fundamental changes. Following up with another PR.

PR #228, opened 2019-10-25
Set speed dial if key is not binded

### LINKED ISSUE #111: Open assing issue to speed dial if used not set speed dial

FrantisekMichalSebestyen:
Set speed dial if key is not binded.

resolves #111


src/actions/trackerActions.js
```diff
@@ -250,7 +250,9 @@ const setJiraIssueFromSpeedDial = (indices: number[], index: string) => (
   getState: GetState
 ) => {
   const { speedDial } = getState()
-  dispatch(setJiraIssue(indices, speedDial[index] || null))
+  if (speedDial[index])
+    dispatch(setJiraIssue(indices, speedDial[index] || null))
+  else dispatch(updateValue(['focus'], { type: FOCUSES.ISSUE, payload: index }))
```
domiin50: Please, use `setFocus` action to have the logic in just one place.

domiin50:
Let's wait to be #227 approved and make this extension after that.

src/actions/trackerActions.js
```diff
@@ -254,7 +254,9 @@ const setJiraIssueFromSpeedDial = (indices: number[], index: string) => (
   getState: GetState
 ) => {
   const { speedDial } = getState()
-  dispatch(setJiraIssue(indices, speedDial[index] || null))
+  if (speedDial[index])
+    dispatch(setJiraIssue(indices, speedDial[index] || null))
```
domiin50:
```suggestion
    dispatch(setJiraIssue(indices, speedDial[index]))
```

PR #235, opened 2019-11-12
Get user schedule from Tempo

FrantisekMichalSebestyen:
Make user endpoint to retrieve schedule of user.
Get information about dates of working and not working days.
Get required working time for every date.


.env.template
```diff
@@ -39,6 +39,8 @@ ft_jira_api_call_myself=https://vacuum.atlassian.net/rest/api/2/myself?expand=gr
 ft_jira_api_call_issues=https://vacuum.atlassian.net/rest/api/3/search?fields=summary
 
 ft_tempo_api_get_worklogs=https://api.tempo.io/core/3/worklogs
+ft_tempo_api_get_worklogs=https://api.tempo.io/core/3/worklogs
```
mikx98: this is the same as the line above :)

server/apiCallbacks.js
```diff
@@ -9,15 +9,38 @@ import { getTempoRequest } from './auth/utils'
...
+  const to = moment()
+    .endOf('month')
+    .format('YYYY-MM-DD')
 
+  const data = await request({
+    method: 'GET',
+    uri: config.tempo.api.userSchedule,
+    qs: {
+      from,
+      to
```
mikx98: feel free to inline the parameters here

mikx98:
@FrantisekMichalSebestyen as discussed, please forward to frontend only necessary data, do the mapping on backend side please :pray:

server/apiCallbacks.js
```diff
@@ -9,15 +9,45 @@ import { getTempoRequest } from './auth/utils'
...
+    json: true
+  })
+  const today = moment().format('YYYY-MM-DD')
+
+  const monthlyRequiredSeconds = data.results
+    .map(day => parseInt(day.requiredSeconds, 10))
+    .reduce((a, b) => a + b, 0)
+
+  const currentlyRequiredSeconds = data.results
+    .map(day => (day.date > today ? 0 : parseInt(day.requiredSeconds, 10)))
```
mikx98: you could have used filter for discarding some data, but no need to change, I like it this way as well :)

server/apiCallbacks.js
```diff
@@ -9,15 +9,45 @@ import { getTempoRequest } from './auth/utils'
...
+    .reduce((a, b) => a + b, 0)
+
+  return { monthlyRequiredSeconds, currentlyRequiredSeconds }
```
domiin50: I would convert it also to hours. On the frontend, we do not want to make other conversions.

FrantisekMichalSebestyen: Tempo is showing this in 15 minutes intervals. So maybe it would be better to convert it to minutes ?

server/apiCallbacks.js
```diff
@@ -35,7 +35,16 @@ const getUserSchedule = async (req, res) => {
     .map(day => (day.date > today ? 0 : parseInt(day.requiredSeconds, 10)))
     .reduce((a, b) => a + b, 0)
 
-  return { monthlyRequiredSeconds, currentlyRequiredSeconds }
+  const monthlyRequiredHours = Math.floor(monthlyRequiredSeconds / 3600)
+  const monthlyRequiredMinutes = Math.floor(
+    (monthlyRequiredSeconds - monthlyRequiredHours * 3600) % 60
+  )
+  const monthlyRequiredString = `${monthlyRequiredHours}:${monthlyRequiredMinutes}`
```
domiin50: Use `moment.duration` and `moment.format` instead of this magic, please :pray: 
Something like: `moment.duration(monthlyRequiredSeconds,  'seconds').format('HH:mm')`

FrantisekMichalSebestyen: The problem with using moment duration is that it reqiuers magic as well. The default behavior of this formatting will not work if the duration is longer than 24 hours. I also have a bad feeling about having state having both duration in seconds and formatted duration. It seems like a bit of antipattern. https://stackoverflow.com/questions/13262621/how-do-i-use-format-on-a-moment-js-duration
Maybe @mikx98 can share his opinion on this one.

mikx98: I wouldn't format data on the BE side at all, that's FE responsibility. I would return json with seconds or minutes or whatever you choose (probably the most granular the best), but only one representation

server/apiCallbacks.js
mikx98: conversions are cheap, it's only one division - let's return the data only once

server/apiCallbacks.js
domiin50: Okay, let's revert the last changes and merge

PR #246, opened 2019-11-13
Add date picker

### LINKED ISSUE #191: Unable to log retrospectively to another day
With the current design I am not able to log retrospectively to another day. For example yesterday.

domiin50:
It is hard to log retrospectively to the past days. It is needed to go
back in time through the midnight to get at previous day.

Add date input with date picker from calendar to easier manipulation
with days of the work logs.

This resolves #191.


domiin50:
Sorry for the long PR :see_no_evil: There are a few libraries changed and added.
The main functionality is in the new _LoggerDateInput_ and in _Logger/index_ files.

src/tracker/Logger/LoggerDateInput.js
```diff
@@ -0,0 +1,100 @@
...
+    disabled:
+      mode === MODES.NORMAL &&
+      firstSelectedPoint &&
+      (selection.length > 1 || firstSelectedPoint.touchesSynced)
+  })),
+  withHandlers({
+    handleArrow: ({ mode, value, setValue, setModified }) => (key: string) => {
+      if (ARROWS[key] > 0 && isToday(value)) return
+      setValue(addDays(value, ARROWS[key]))
+      setModified(mode === MODES.LOGGING)
```
mikx98: why are we not setting it as modified in normal mode?

src/tracker/Logger/LoggerDateInput.js
```diff
@@ -0,0 +1,100 @@
...
+    ) => {
+      setValue(date.toDate())
+      setModified(mode === MODES.LOGGING)
+      mode === MODES.NORMAL && handleSubmit(date)
+    }
+  }),
+  withHandlers({
+    handleKeyDown: ({ handleSubmit, handleArrow }) => (e: any) => {
+      const keyBindings = {
+        Enter: () => handleSubmit(),
```
mikx98:
```suggestion
        Enter: handleSubmit,
```

src/tracker/Logger/LoggerDateInput.js
domiin50: Because this property is used for resetting the timestamp after focus when it has not been changed in the logging mode. In the normal mode, we do not need this functionality. We do not want to reset the timestamp to the actual one.

PR #248, opened 2019-11-13
Create HOC for disabling property

domiin50:
The same disabling functionality is duplicated at more different places.

Move inputs disabling logic into one HOC that is imported for getting
disabled property.

Rename the all HOC variables to preserve the convention of
functions namings.


src/tracker/Logger/withDisabling.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+import { connect } from 'react-redux'
+import { compose, withProps } from 'recompose'
+
+import firstSelectedPointSelector from '../firstSelectedPointSelector'
+import { MODES } from '../../consts'
+
+const Disabling = compose(
```
mikx98: this is not a component by itself, please use `withDisabling` as the variable name here - HOCs are only functions

src/tracker/Logger/LoggerDateInput.js
```diff
@@ -62,17 +62,7 @@ const LoggerDateInput = ({
...
-  connect(state => ({
-    mode: state.mode,
-    selection: state.selection,
-    firstSelectedPoint: firstSelectedPointSelector(state)
-  })),
-  withProps(({ mode, selection, firstSelectedPoint }) => ({
-    disabled:
-      mode === MODES.NORMAL &&
-      firstSelectedPoint &&
-      (selection.length > 1 || firstSelectedPoint.touchesSynced)
```
mikx98: you also added `|| mode === MODES.SAVING` here (which is correct, but please mention such things in PR description next time)

domiin50: I did not add it. It was just moved from the TimeInput.

PR #250, opened 2019-11-13
Reverse worklogs in overview

### LINKED ISSUE #225: Time from older on top to the newest on the bottom in new design
The order of agenda looks contra-intuitive to me, I would prefer to have it in reverse - to have newest thing on the bottom, right above the input field. 

Let's discuss this with @darjanin

domiin50:
Make worklogs agenda more intuitive. Reverse the worklogs to have the
newest one on the bottom and the oldest one on the top.

This resolves #225.


src/tracker/WorkLogs/worklogs.css
```diff
@@ -26,15 +26,15 @@
 }
 
 .worklogs > .container {
-  height: calc(100% - 69px);
-  align-items: center;
-  padding-bottom: 15px;
+  height: calc(100% - 54px);
   overflow-y: auto;
+  display: flex;
+  flex-direction: column-reverse;
```
FrantisekMichalSebestyen: Scrolling is not working on some browsers.
Can you please have a look ?

FrantisekMichalSebestyen: https://github.com/philipwalton/flexbugs/issues/108

PR #252, opened 2019-11-15
Change header date according to selected log

### LINKED ISSUE #245: Change Timeline and header date according to the currently selected log
Closing #230 #231 and #232 as decided on design meeting.
Timeline and displayed date will change according to the currently selected log.

_Originally posted by @FrantisekMichalSebestyen in https://github.com/vacuumlabs/fine-tracker/pull/232#issuecomment-552516418_

FrantisekMichalSebestyen:
Change date in timeline header and timeline to currently selected worklog.

Resolves #245


src/tracker/displayedDateSelector.js
```diff
@@ -0,0 +1,16 @@
...
+import { createSelector } from 'reselect'
+import type { State } from '../store/state'
+
+const pointsSelector = (state: State) => state.points
+const cursorSelector = (state: State) => state.cursor
+
+const displayedDateSelector = createSelector(
+  [pointsSelector, cursorSelector],
+  (points, cursor) =>
+    points[cursor.index] ? points[cursor.index].timestamp : moment()
```
domiin50: What about just a classic new Date(). Let's use JS basics if it is not needed otherwise.

domiin50: I would probably use in the condition also the actual mode.
Do not know yet what do we want to display in the timeline in `Saving mode`.
But in `Logging mode` there should be probably an actual date.

src/tracker/TimelineHeader/index.js
```diff
@@ -12,6 +12,7 @@ import './timelineHeader.css'
 
 import { switchMode as _switchMode } from '../../actions/trackerActions'
 import { MODES } from '../../consts'
+import displayedDateSelector from '../displayedDateSelector'
```
domiin50: Just thinking :thinking: What about to use `firstSelectedPointSelector`?
Do we want to change the date and timeline also when there is a selection and moving cursor? The input values are not changing in this situation.
I probably like it more this way as it is done, but the `firstSelectedPointSelector` is also a possibility.
Wdyt? Also @mikx98?

mikx98: as discussed on standup, I lean more towards using `firstSelectedPointSelector` here

src/tracker/displayedDateSelector.js
mikx98: I wouldn't change timeline view when switching `Normal mode` <-> `Saving mode` - keep there the same thing

mikx98: I agree with the `Logging mode` - let's put there the current day. maybe longer term goal would be to show the day that is selected in the date input? :thinking: but don't think about that one now :)

domiin50: But in `Saving mode` there are is no point returned as `firstSelected` from the selector. The timeline is totally worthless here in this mode IMO. So I suggest putting there also the current day as in `Logging mode` meantime.

src/tracker/displayedDateSelector.js
```diff
@@ -1,16 +1,14 @@
...
-const pointsSelector = (state: State) => state.points
-const cursorSelector = (state: State) => state.cursor
+const modeSelector = (state: State) => state.mode
 
 const displayedDateSelector = createSelector(
-  [pointsSelector, cursorSelector],
-  (points, cursor) =>
-    points[cursor.index] ? points[cursor.index].timestamp : moment()
+  [firstSelectedPointSelector, modeSelector],
+  (point, mode) => (point ? point.timestamp : new Date())
```
domiin50: `mode` is not needed here I guess since it is handled in `firstSelectedPointSelector` already

src/tracker/displayedDateSelector.js
mikx98: hmm, my point was to not hide & show the timeline when entering & leaving the saving mode, as it may be quite confusing ... maybe we want to modify the selector?

src/tracker/displayedDateSelector.js
domiin50:
```suggestion
  point => point ? point.timestamp : new Date()
```

PR #254, opened 2019-11-15
Disable mode switching if issues are not loaded

### LINKED ISSUE #253: Disable saving mode before issues are loaded
If issues are not loaded before entering saving mode FineTracker throws an error.
Saving mode should be disabled before they are loaded

FrantisekMichalSebestyen:
Fix error throwing if issues are not loaded and user attempts to switch modes
Resolves #253


src/actions/trackerActions.js
```diff
@@ -222,7 +222,10 @@ export const switchMode = (mode: Mode) => (
   dispatch: Dispatch,
   getState: GetState
 ) => {
-  const editablePoints = editablePointsSelector(getState())
+  const state = getState()
+  const editablePoints = editablePointsSelector(state)
+  const issues = state.issues
+  if (issues.length < 1) return
```
mikx98: I don't think that we want to disable switching modes when there are no issues loaded - please handle this case elsewhere. For example in that component where the error is happening show something like - `Receiving issues from Jira, wait a moment please` or something similar

FrantisekMichalSebestyen:
Issues is resolved in other way. I'm closing this PR and creating new one.

PR #255, opened 2019-11-18
Move through already submitted log

### LINKED ISSUE #244: Move cursor through already saved worklogs
Right now the cursor is skipping through worklogs already saved to JIRA.
This is not the behaviour we want.
Make cursor able to move through already saved worklogs.

FrantisekMichalSebestyen:
Make user able to scroll through submitted logs.
Delete scrolling from top to bottom and vice versa.
Resolves #244


src/actions/trackerActions.js
```diff
@@ -178,31 +177,22 @@ export const setCursor = (index: number) => (
...
-  const [newEditableIndex] = [cursor.editableIndex + increment]
-    .map(acc => (acc === -1 ? points.length - 1 : acc))
-    .map(acc => (acc === points.length ? 0 : acc))
-    .map(acc => clamp(acc, 0, points.length - 1))
-
-  const newCursor = {
-    editableIndex: newEditableIndex,
-    index: points[newEditableIndex].index
-  }
+  const { cursor, points } = state
```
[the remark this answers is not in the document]
FrantisekMichalSebestyen: Just out of curiosity. Do you know whether this is more effective, than previous solution ?

FrantisekMichalSebestyen:
I need help with disabling Issue input. Is there a way to send payload prop to Component. So I can disable it only when payload is missing ?

domiin50:
@FrantisekMichalSebestyen, does this work? Is it ready for a review, or is it still blocked?

mikx98:
@FrantisekMichalSebestyen some bugs I discovered in a brief testing:
1. If I go down when I am on the last worklog, I lose focus somewhere. Same if I go up on the first worklog. We don't want that
2. If I am on a synced worklog and press "6" (i don't have a speed dial option assigned), issue input gets enabled. If I press it again, I am assigning the speed dial. Also, shortcut `S` does something.

src/actions/trackerActions.js
```diff
@@ -183,12 +182,10 @@ export const setCursor = (index: number) => (
...
   getState: GetState
 ) => {
-  const editablePoints = editablePointsSelector(getState())
-  !editablePoints.length ||
-    (index < 0 && dispatch(updateValue(['cursor'], getInitialCursorState())))
-  const editableIndex = editablePoints.findIndex(point => point.index === index)
-  editableIndex >= 0 &&
-    dispatch(updateValue(['cursor'], { editableIndex, index }))
+  const points = getState().points
+  points.length < 1 || index < 0
```
mikx98: what is the purpose of this condition? I would say that this function should only ensure that the cursor is between `0 <= index < points.length` + we may have a special handling for `index=-1` (which is unclear to me now, what the semantics of -1 cursor is). do we actually need it?

src/actions/trackerActions.js
```diff
@@ -206,19 +203,10 @@ export const moveCursor = (increment: number) => (
...
-    .map(acc => clamp(acc, 0, points.length - 1))
-
-  const newCursor = {
-    editableIndex: newEditableIndex,
-    index: points[newEditableIndex].index
-  }
+  const { cursor, points } = getState()
+  const [newCursor] = [cursor + increment]
+    .map(acc => (acc === -2 ? -1 : acc))
+    .map(acc => (acc === points.length + 1 ? points.length : acc))
```
mikx98: only using clamp is enough here (instead of those 2 maps)

src/actions/trackerActions.js
```diff
@@ -230,21 +218,18 @@ export const toggleSelectCursorPoint = () => (
...
 ) => {
-  const editablePoints = editablePointsSelector(getState())
+  const { points } = getState()
   if (mode === MODES.NORMAL) {
     blurElement()
-    const newCursor = editablePoints.length
-      ? editablePoints[editablePoints.length - 1].index
-      : -1
-    dispatch(setCursor(newCursor))
+    dispatch(setCursor(points.length))
```
mikx98: maybe this?: 
```suggestion
    dispatch(setCursor(points.length - 1))
```

src/actions/trackerActions.js
```diff
@@ -323,9 +308,12 @@ const makeSelectionAction = actionCreator => (...args: any[]) => (
...
   getState: GetState
 ) => {
-  const { cursor, selection } = getState()
+  const { cursor, selection, points } = getState()
   dispatch(
-    actionCreator(selection.length ? selection : [cursor.index], ...args)
+    actionCreator(
+      selection.length ? selection : points[cursor].synced ? [] : [cursor],
+      ...args
+    )
```
mikx98: :+1:

src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -56,26 +58,27 @@ const IssueInput = ({
...
+      placeholder="Search JIRA issues"
+      onKeyDown={handleKeyDown}
+      onFocus={handleFocus}
+      onChange={handleChange}
+      onBlur={handleBlur}
+      disabled={disabled}
+      inputRef={handleRef}
+    />
+    {mode === MODES.NORMAL && <kbd>S</kbd>}
+  </div>
```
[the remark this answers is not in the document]
FrantisekMichalSebestyen: Could you please clarify ? thx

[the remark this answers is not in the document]
mikx98: check the github's diff, you removed 1 tab from this code block, but it should be there :)

mikx98:
@FrantisekMichalSebestyen please do this one as a priority, we want to merge this as soon as possible :D finally :)

src/tracker/Logger/withDisablingDescription.js
```diff
@@ -5,19 +5,19 @@ import { compose, withProps } from 'recompose'
...
     selection: state.selection,
+    points: state.points,
     firstSelectedPoint: firstSelectedPointSelector(state)
   })),
   withProps(({ mode, selection, firstSelectedPoint }) => ({
     disabled:
       (mode === MODES.NORMAL &&
-        firstSelectedPoint &&
-        (selection.length > 1 || firstSelectedPoint.touchesSynced)) ||
+        (firstSelectedPoint && firstSelectedPoint.synced)) ||
```
mikx98: I think all stuff should be disabled also if `firstSelectedPoint` is null (empty selection in normal mode -> no edit action possible)

src/tracker/JiraSidebar/withDisablingIssueInput.js
```diff
@@ -0,0 +1,28 @@
...
+    mode: state.mode,
+    points: state.points
+  })),
+  withProps(({ selection, firstSelectedPoint, mode, points, payload }) => {
+    if (payload) return { disabled: false }
+    if (mode === MODES.NORMAL && firstSelectedPoint)
+      return {
+        disabled: selection.length
+          ? some(selection, index => points[index].description === '')
+          : firstSelectedPoint.description === '' || firstSelectedPoint.synced
```
mikx98: any selected worklog (if multiple) can be synced and we have to disable

mikx98: a nice solution would be to create a selector for `selection: selection.length ? selection : [cursor]` and write this condition as `disabled: selection.some(..........)`

src/actions/trackerActions.js
```diff
@@ -249,9 +227,16 @@ const makeSelectionAction = actionCreator => (...args: any[]) => (
...
-  const { cursor, selection } = getState()
+  const { cursor, selection, points } = getState()
   dispatch(
-    actionCreator(selection.length ? selection : [cursor.index], ...args)
+    actionCreator(
+      selection.length
+        ? selection
+        : points[cursor] && points[cursor].synced
+        ? []
+        : [cursor],
```
mikx98: please don't use double ternar operator :pray: `a ? b : c ? d : e`, it's confusing. also once you create the selector mentioned in other comment, you can use it here

PR #256, opened 2019-11-18
Show loading icon, if issues are not loaded

### LINKED ISSUE #253: Disable saving mode before issues are loaded

FrantisekMichalSebestyen:
If issues are not fetched from JIRA, FineTracker throws an error.
Instead show Loading icon.
This was resolved in Issues #254 in bad way.

Resolves #253


src/tracker/WorklogsToSave/index.js
```diff
@@ -45,10 +46,11 @@ const ShowStateDialog = ({ state: { loading, error, success } }) => (
 type Props = {|
   handleSubmit: () => void,
   switchMode: (mode: Mode) => void,
-  classes: Object
+  classes: Object,
```
domiin50: This prop is not really in the object. Feel free to remove it.

src/tracker/WorklogsToSave/index.js
```diff
@@ -61,22 +62,28 @@ const WorklogsToSave = ({ handleSubmit, switchMode }: Props) => (
...
+    )}
   </div>
 )
 
 export default compose(
   connect(
-    null,
+    state => ({ issues: state.issues }),
     { submitWorklogs: _submitWorklogs, switchMode: _switchMode }
   ),
```
mikx98: you can write here:
```suggestion
  branch(
    ({ issues }) => !issues.length,
    renderComponent(Loading)
  )
```
and avoid this logic in the component itself :) imo it's nicer this way

FrantisekMichalSebestyen: Using branch is nice, but look kinda weird to me, when shortcuts panel moves so much.

PR #257, opened 2019-11-20
Return to saving mode after error

### LINKED ISSUE #236: Return to saving mode after unsuccessful submit
Show worklogs while saving issues to tempo.
If saving is successful go back to normal, else stay in saving mode and refresh the issues.

FrantisekMichalSebestyen:
Return to saving mode, after unsuccessful submit 
Show saving mode in background, while saving.

This resolves #236


src/tracker/WorklogsToSave/index.js
```diff
@@ -41,11 +36,13 @@ const ShowStateDialog = ({ state: { loading, error, success } }) => (
 type Props = {|
   handleSubmit: () => void,
   switchMode: (mode: Mode) => void,
-  classes: Object
+  classes: Object,
+  state: Object
 |}
 
-const WorklogsToSave = ({ handleSubmit, switchMode }: Props) => (
+const WorklogsToSave = ({ handleSubmit, switchMode, state }: Props) => (
```
domiin50: I know, that you did not change the name. But please, could you rename this prop? It is misleading with the Redux state.

PR #264, opened 2019-12-09
Add Date Splitter

### LINKED ISSUE #241: Move the date of worklogs to the gap between days
Enhance the gap in between days.
Move the date of worklogs to the gap.

domiin50:
Display day transition another way.

Remove date from the end timestamp of worklogs.

Add date splitter element to split days in the agenda.

It resolves #241.


domiin50:
> I was expecting the date to be a bit smaller

Do we want to ask for a mockup? Or how should it look like? @mikx98

src/tracker/WorkLogs/index.js
```diff
@@ -49,7 +58,7 @@ export default compose(
   connect(
     state => ({
       mode: state.mode,
-      events: eventsSelector(state)
+      eventsByDate: workLogsEventsSelector(state)
```
mikx98: I would rename the selector to something similar to `eventsByDateSelector` - name according to what it does instead of where it's used

domiin50:
> imho it looks good, but I would make the date splitters at least half the height (it's only additional information)
> 
> also, I got a feeling from the design meeting, that if you have a work started on Monday and next log would be on Wednesday, you would also have the Tuesday day splitter there. Not sure about this, though. You can try raising in #fine-tracker-dev channel, maybe @jakuub remembers better

According to Slack conversation, this will be done in separate PR.

PR #266, opened 2019-12-10
Extract Speeddial options and Recent Searches from Jira Sidebar

FrantisekMichalSebestyen:
Prepare for future modification in Jira sidebar component


src/tracker/JiraSidebar/RecentSearches.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+import React from 'react'
+import { connect } from 'react-redux'
+import { compose } from 'recompose'
+import bind from 'memoize-bind'
+import { sidebarRecentIssuesSelector } from '../issuesSelectors'
+import type { Issue } from '../../store/state'
+
+import './sidebar.css'
```
domiin50: No need for importing the css more than once in the whole project.
It is already imported in the `index`, I guess.

src/tracker/JiraSidebar/RecentSearches.js
```diff
@@ -0,0 +1,42 @@
...
+      ))}
+    </div>
+  </React.Fragment>
+)
+export default compose(
+  connect(
+    state => ({
+      recentIssues: sidebarRecentIssuesSelector(state)
+    }),
+    {}
```
domiin50: The `connect` does not need an empty object here. The second parameter will be empty if not defined.

src/tracker/JiraSidebar/SpeedDialOptions.js
```diff
@@ -0,0 +1,48 @@
...
+import { connect } from 'react-redux'
+import { compose } from 'recompose'
+import bind from 'memoize-bind'
+
+import type { SpeedDial } from './speedDialSelector'
+import speedDialSelector from './speedDialSelector'
+
+import { setJiraIssueFromSpeedDialForSelection as _setJiraIssueFromSpeedDial } from '../../actions/trackerActions'
+
+import './sidebar.css'
```
domiin50: Same. Please, remove.

PR #267, opened 2019-12-10
Move issue input to the bottom of sidebar

FrantisekMichalSebestyen:
Move issue input to the the bottom of JIRA Sidebar
Make issue options appear in column above issue input


src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -18,50 +16,38 @@ import {
 
 import { blurElement } from '../../utils'
 
-import {
-  inputIssuesSelector,
-  inputRecentIssuesSelector,
-  allIssuesObjSelector,
-  mapIssuesToInputOptions
-} from '../issuesSelectors'
+import { inputIssuesSelector, allIssuesObjSelector } from '../issuesSelectors'
```
domiin50: Since we do not need a special form of issue object for the async input anymore, you can remove all the selectors that map normal issues to the specific format for late input.

It is no needed `inputIssuesSelector` here. Just normal `allIssuesSelector` selector.

src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -1,12 +1,10 @@
 // @flow
 import React from 'react'
 import { connect } from 'react-redux'
-import { compose, withState, withHandlers } from 'recompose'
+import { compose, withHandlers, withState } from 'recompose'
 import { withTheme } from '@material-ui/core/styles'
 
-import debounce from 'es6-promise-debounce'
-
-import AsyncSelect from 'react-select/lib/Async'
```
domiin50: You can remove the unused libs also from `package.json`.

src/tracker/JiraSidebar/index.js
```diff
@@ -65,6 +93,8 @@ export default compose(
       switchMode: _switchMode
     }
   ),
+  withState('input', 'setInput', null),
```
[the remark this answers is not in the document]
FrantisekMichalSebestyen: Probably doesn't matter a lot, but i didn't want to name it `issue` because it doesn't have to be issue what is written there. Maybe `issueInput` is better ?

src/tracker/JiraSidebar/IssueOptions.js
```diff
@@ -0,0 +1,31 @@
...
+import { map } from 'lodash'
+
+import type { Issue } from '../../store/state'
+
+type Props = {|
+  issueOptions: Issue[],
+  handleIssueClick: (value: string) => void
+|}
+
+const IssuesOptions = ({ issueOptions, handleIssueClick }: Props) => (
```
domiin50: Please, keep the namings the same as the name of the file.
`IssuesOptions` -> `IssueOptions`

src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -83,58 +69,61 @@ export default compose(
...
+      const regexes = issueInput
+        .split(' ')
+        .map(value => new RegExp(`\\b${value}`, 'i'))
+
+      setIssueInput(issueInput)
+      setIssueOptions(
+        filter(allIssuesObj, issue =>
+          regexes.every(regex => regex.test(issue.key))
+        ).slice(0, 20)
+      )
```
domiin50: `IssueOptions` should be just filtered `allIssuesObj` object? Or the filtered array of the Issues? Because here is inconsistent typing of this prop.
It is needed to make clear, in what form should be the issues in this property and so, which selector use and how to access the issues in it.

PR #271, opened 2019-12-11
Scroll last issue option into view

FrantisekMichalSebestyen:


FrantisekMichalSebestyen:
Left it there because It is an improvement to the current situation

PR #274, opened 2019-12-20
Add copy worklog shortcut

### LINKED ISSUE #262: Re-log focused log to current time
When in normal mode, hovering over some worklog (cursor on that worklog), we could have some fancy shortcut (f.e. `R`) that would take the same log (description + issue), give it current time and log it now.

FrantisekMichalSebestyen:
Add optional issue parameter to createPoint
Add copy worklog functionality to r key shortcut

This resolves #262


src/actions/trackerActions.js
```diff
@@ -14,9 +14,13 @@ import type { Dispatch, GetState } from '../types/reduxTypes'
 
 import { PREHISTORIC_DATE, MODES, FOCUSES } from '../consts'
 
-export const createPoint = (description: string, timestamp: string) => ({
+export const createPoint = (
+  description: string,
+  timestamp: string,
+  issue?: ?string
```
mikx98: to be better backward compatible, we want to stay away from setting undefined as the issue. Hence issue is either supplied, or it's null. 
```suggestion
  issue?: ?string = null
```

src/tracker/keyBindingsUtils.js
```diff
@@ -123,6 +125,11 @@ const getKeyBindings = ({
           handler: () => switchMode(MODES.LOGGING),
           category: 'Creating'
         },
+        r: {
+          description: 'Copy worklog',
+          handler: () => copyWorklog(),
```
mikx98:
```suggestion
          handler: copyWorklog,
```

src/actions/trackerActions.js
```diff
@@ -186,6 +190,17 @@ export const setCursor = (index: number) => (
     dispatch(updateValue(['cursor'], { editableIndex, index }))
 }
 
+export const copyWorklog = () => (dispatch: Dispatch, getState: GetState) => {
+  const { cursor, points } = getState()
+  dispatch(
+    createPoint(
+      points[cursor.index].description,
```
mikx98: hm, I am thinking that in order to be consistent with other shortcuts' behavior, we should consider selection as well. probably use the firstSelectedPoint selector and create point based on that point's description and issue?

PR #298, opened 2020-01-17
Sort issues based on recent searches

### LINKED ISSUE #276: Order filtered issues based on most recent usage
It was once implemented, #44, but disappeared in the git history.

FrantisekMichalSebestyen:
Add recent issues selector

Resolves #276


src/tracker/issuesSelectors.js
```diff
@@ -31,3 +31,12 @@ export const sidebarRecentIssuesSelector: InputSelector<
   recentIssuesSelector,
   issues => issues.slice(0, 3)
 )
+
+export const inputRecentIssuesSelector: InputSelector<
```
mikx98: this doesn't return only recent issues, please rename to something like: `inputIssuesSelector` (based on where it's used) or better yet, to `allIssuesWithRecentFirstSelector`

src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -93,8 +93,8 @@ export default compose(
       allIssues
     }) => () => {
       setIssueInput('')
-      setIssueOptions(allIssues)
-      setCursor(allIssues.length - 1)
+      setIssueOptions(reverse(allIssues.slice(0, 20)))
+      setCursor(19)
```
mikx98: constant 20 is used 3 times in this file now. please make it a constant for easier manipulation later

src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -93,8 +94,8 @@ export default compose(
       allIssues
     }) => () => {
       setIssueInput('')
-      setIssueOptions(allIssues)
-      setCursor(allIssues.length - 1)
+      setIssueOptions(reverse(allIssues.slice(0, ISSUE_OPTIONS_LENGTH)))
+      setCursor(ISSUE_OPTIONS_LENGTH - 1)
```
mikx98: hm, this will probably never happen, but in case there are only 5 issues, we will set the cursor to 19 still

src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -104,9 +105,13 @@ export default compose(
           .map(value => new RegExp(`\\b${value}`, 'i'))
         return new Promise(resolve =>
           resolve(
-            filter(allIssues, issue =>
-              regexes.every(regex => regex.test(`${issue.title} ${issue.key}`))
-            ).slice(0, 20)
+            reverse(
+              filter(allIssues, issue =>
+                regexes.every(regex =>
+                  regex.test(`${issue.title} ${issue.key}`)
```
mikx98: we can maybe use the issueLabelFormatting util function from the other PR here as well, wdyt?

src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -20,12 +20,13 @@ import {
...
 import { blurElement } from '../../utils'
 
-import { allIssuesSelector } from '../issuesSelectors'
+import { allIssuesWithRecentFirstSelector } from '../issuesSelectors'
 
 import { FOCUSES, MODES } from '../../consts'
 import type { Mode } from '../../consts'
 import type { Issue, Point } from '../../store/state'
 
+const ISSUE_OPTIONS_LENGTH = 20
```
mikx98:
```suggestion
const ISSUE_OPTIONS_MAX_LENGTH = 20
```

src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -93,8 +94,8 @@ export default compose(
       allIssues
     }) => () => {
       setIssueInput('')
-      setIssueOptions(allIssues)
-      setCursor(allIssues.length - 1)
+      setIssueOptions(reverse(allIssues.slice(0, ISSUE_OPTIONS_LENGTH)))
```
mikx98: did you test this? you renamed the variable, but in the code it is not renamed :cry: :scream:

src/tracker/JiraSidebar/IssueInput.js
```diff
@@ -96,20 +100,28 @@ export default compose(
       allIssues
     }) => () => {
       setIssueInput('')
-      setIssueOptions(allIssues)
-      setCursor(allIssues.length - 1)
+      setIssueOptions(reverse(allIssues.slice(0, ISSUE_OPTIONS_MAX_LENGTH)))
+      setCursor(
+        allIssues.length < 20
+          ? allIssues.length - 1
+          : ISSUE_OPTIONS_MAX_LENGTH - 1
```
mikx98: :) just save the `reverse(allIssues.slice(0, ISSUE_OPTIONS_MAX_LENGTH))` into some variable (e.g. `trimmedResults`) and use that lentgth's -1 here

PR #315, opened 2020-01-28
Help with setting payload

FrantisekMichalSebestyen:
I want to set payload on description input.
But it's not set when focusing the object.


FrantisekMichalSebestyen:
Resolved by using timeout

PR #321, opened 2020-01-30
Add insert function

### LINKED ISSUE #289: Add back insert function
Let's add shortcut `i` which will focust insertin of new log, but leave pre-filled date and time from current point.

FrantisekMichalSebestyen:
Resolves #289


src/consts.js
```diff
@@ -3,6 +3,7 @@ export type Mode = 'logging' | 'saving' | 'normal'
 
 export type FocusType = 'description' | 'time' | 'issue'
 export type DialogType = 'joinIntervals' | 'submitWorklogs'
+export type FocusPayloadType = string | 'insert'
```
mikx98: what do we use on assign speed dial option? don't we want to include it here? name it please

src/actions/trackerActions.js
```diff
@@ -254,8 +254,7 @@ const setJiraIssueFromSpeedDial = (indices: number[], index: string) => (
   getState: GetState
 ) => {
   const { speedDial } = getState()
-  if (speedDial[index])
-    dispatch(setJiraIssue(indices, speedDial[index]))
+  if (speedDial[index]) dispatch(setJiraIssue(indices, speedDial[index]))
   else dispatch(setFocus(FOCUSES.ISSUE, index))
```
mikx98: index is not focus payload TYPE. please do it a little differently, to be consistent with the naming

src/consts.js
```diff
@@ -16,6 +17,10 @@ export const FOCUSES = {
   ISSUE: 'issue'
 }
 
+export const FOCUS_PAYLOAD_TYPES = {
+  INSERT: 'insert'
```
mikx98: I think `insert` is not that much descriptive :/

src/tracker/Logger/LoggerInput.js
```diff
@@ -83,7 +90,30 @@ export default compose(
...
+            : new Date()
+        switchMode(MODES.LOGGING)
+        setTimeout(() => {
+          setTimestamp(new Date(timestamp))
+        })
+        setPayload(null)
+        return
+      }
+      handleFocus()
+    },
```
mikx98: ah jeez, we really need to refactor this payload sending functionality :cry: :scream: but good for this PR

src/actions/trackerActions.js
FrantisekMichalSebestyen: @mikx98 index is a string. Could you please help ?

mikx98: discussed on standup

PR #324, opened 2020-02-03
Set points as synced one by one

### LINKED ISSUE #297: Mark intervals as synced one by one instead of all at once
When I have tried to sync 2 weeks worth of logs (yes shame on me) to JIRA it took ~20 seconds. Being impatient I have refreshed the page after a few seconds and ended up with nothing marked as synced, but most of my logs being in JIRA.

Every `submit` request should mark the respective interval as synced, otherwise is very easy to end up with things logged multiple times as a result of network error / impatience.

FrantisekMichalSebestyen:
Resolves #297


src/actions/trackerActions.js
```diff
@@ -291,14 +291,15 @@ export const submitWorklogs = () => async (
...
+        dispatch(setPointAsSynced(data.id))
+        return data.id
+      })
   )
 
-  const responses = await Promise.all(promises)
-  const syncedIds = (await Promise.all(responses.map(res => res.json())))
-    .map(r => r.id)
-    .filter(id => id >= 0)
+  const syncedIds = (await Promise.all(promises)).filter(id => id >= 0)
```
mikx98: we don't need this variable, so we also don't need the filtering here (do it above, in the second `then`)
this line can be replaced by this:
```suggestion
  await Promise.all(promises))
```

FrantisekMichalSebestyen: @mikx98 How should I compute the number of succesful promises ?

mikx98: aaaah, I thought that was the end of the function as github showed :smile: sorry, I see that syncedIds is actually used and you need it. forget my comment :)

PR #326, opened 2020-02-04
Add error message when trying to affect synced worklogs

### LINKED ISSUE #285: Error message on not applied change because of protected intervals
When someone changes something (e.g. add a point) to already submitted interval, change is not applied. It will be great to show some error, for the user to know, why the change was not applied.

FrantisekMichalSebestyen:
Add snackbar displaying error message
Turn actions editing worklogs into pure functions
Create middleware that recognizes wheter user is
trying to modify synced worklogs

Closes #285


src/actions/trackerActions.js
```diff
@@ -330,11 +261,36 @@ const makeSelectionAction = actionCreator => (...args: any[]) => (
   dispatch(clearSelection())
 }
 
-export const deleteSelection = makeSelectionAction(deletePoints)
+const makeEditPointsAction = (editPointsFunction: Function) => (
```
mikx98: If you chose to outsource all the points handler functions into a separate file, I would suggest to outsource this helper there too and export already functions bound with this decorator. The output would be that this file here will not know anything about `makeEditPointsAction` decorator, it doesn't have to

FrantisekMichalSebestyen: Hmm... Where should I put setErrorMessage ? I don't want to create 2-way dependency. Should I create separate file for that ?

FrantisekMichalSebestyen: @mikx98  Maybe we should also export makeSelectionAction and import functions from there

mikx98: yes, please create a separate file for general actions and put there the setErrorMessage action

mikx98: and yes, we can put in that _bulkWorklogActions_ file also the makeSelectionAction function and export the actions already all decorated

PR #336, opened 2020-02-12
Add div on disabled buttons

FrantisekMichalSebestyen:
Disabled buttons cannot fire events.
Put div over disabled buttons to toggle tooltips


FrantisekMichalSebestyen:
Hide tooltip on disabled button

src/tracker/Header/index.js
```diff
@@ -118,5 +120,10 @@ export default compose(
       shortcutsShown: state.shortcutsShown
     }),
     { toggleShortcuts: _toggleShortcuts, undo: _undo, redo: _redo }
-  )
+  ),
+  withProps(({ history }) => ({
+    disableUndo: history.index < 1,
+    disableRedo:
```
mikx98: use selectors for these props :pray:

src/tracker/Header/index.js
```diff
@@ -66,21 +70,19 @@ const Header = ({
     <AppBar color="transparent" position="static">
       <Toolbar>
         <img src={Logo} alt="logo" className={classes.logo} />
-        <Tooltip title="Undo">
-          <IconButton onClick={undo} disabled={history.index < 1}>
-            <ArrowBackIosIcon />
-          </IconButton>
+        <Tooltip title={disableUndo ? '' : 'Undo'} placement="top">
```
mikx98: I would prefer to not render the Tooltip component in case it should be disabled at all. Maybe also try to extract these components into variables in render function, I think it can be pretty nice that way :)

PR #341, opened 2020-02-14
Fix error when moving up and down on empty worklog list

FrantisekMichalSebestyen:


mikx98:
as agreed on standup, let's solve this directly in https://github.com/vacuumlabs/fine-tracker/pull/255, as the whole logic changes there

PR #342, opened 2020-02-14
Set timestamp on time input every minute

### LINKED ISSUE #338: Wrong logged time
I am not sure about the exact repro but it is something like:

* Go to logging mode
* Enter some issue
* Wait a lot, probably in background tab
* Enter some new issue *without* touching time field

Expected behavior
* issue is logged with current time

Actual behavior:
* issue is logged with previous time

FrantisekMichalSebestyen:
Resolves #338


src/tracker/Logger/index.js
```diff
@@ -121,7 +121,9 @@ export default compose(
       setTimestamp
     }) => () => {
       mode === MODES.NORMAL && !firstSelectedPoint && switchMode(MODES.LOGGING)
-      mode === MODES.LOGGING && !isTimestampModified && setTimestamp(new Date())
+      mode === MODES.LOGGING &&
+        !isTimestampModified &&
+        setInterval(setTimestamp(new Date()), 60 * 1000)
```
mikx98: you also need a destructor on setInterval :/ it may happen that it rerenders 30 times and there would be 30 distinct set intervals still running, every minute :)

mikx98:
as agreed on tech & normal standups, let's close this PR and solve the issue differently (by showing `now` instead of the changing time)

PR #344, opened 2020-02-17
Show number of hours logged per day on timeline header

### LINKED ISSUE #334: Display today's total number of hours logged
It would be nice to see **today's total number of hours logged** somewhere near the timeline
image

Similar to what we get in Tempo:
image

FrantisekMichalSebestyen:
Create timeline selectors file
Move time formatting to utils
Resolves #334


src/tracker/TimelineHeader/timelineHeader.css
```diff
@@ -5,6 +5,10 @@
   padding-right: var(--PADDING-RIGHT);
 }
 
+.timeline-header > .hours-logged-per-day {
+  padding-left: 15px;
```
mikx98: as discussed on standup, please align to the right, with flexbox styling

src/tracker/eventsSelector.js
```diff
@@ -49,7 +50,15 @@ const eventsSelector: InputSelector<State, any, Event[]> = createSelector(
         actual: isActual,
         hasCursor: isNormalMode && i === cursor,
         selected: isNormalMode && selection.includes(i),
-        synced: point.synced
+        synced: point.synced,
+        timeSpentSeconds:
```
mikx98:
```suggestion
        timeSpentSeconds: i < points.length-1 ? moment.duration(eventEnd.diff(eventStart)).asSeconds() : 0
```

src/tracker/timelineSelectors.js
```diff
@@ -0,0 +1,49 @@
...
+      : [
+          {
+            ...events[events.length - 1],
+            start: moment(displayedDate).startOf('day')
+          }
+        ]
+  }
+)
+
+export const hoursLoggedPerDaySelector: InputSelector<
```
mikx98: this doesn't return hours logged per day, this only returns that on one day - the selected one. Since we are already in file named `timelineSelectors`, I think it's enough to just name like this:
```suggestion
export const hoursLoggedSelector: InputSelector<
```

src/tracker/timelineSelectors.js
```diff
@@ -0,0 +1,49 @@
...
+  State,
+  any,
+  number
+> = createSelector(
+  [timelineEventsSelector],
+  timelineEvents =>
+    formatTime(
+      reduce(
+        timelineEvents,
+        (acc, event) => acc + (event.synced ? event.timeSpentSeconds : 0),
```
mikx98: we discussed some time ago, that this approach may be confusing. we agreed on something like:
1. compute synced worklog time and total (synced + not-synced) time
2. show the total time
3. if the total time is not equal to the synced time (a save action is needed), show the time in red.

@jakuub correct me if I am wrong

mikx98: as discussed, add a comment explaining what the time means

PR #346, opened 2020-02-17
Create separate endpoint for accessing user schedule

FrantisekMichalSebestyen:


src/App.js
```diff
@@ -101,6 +101,12 @@ export default compose(
       ]),
       errorComponent: <Login />
     },
+    {
+      ref: 'schedule',
+      getData: [fetch, '/api/schedule', { credentials: 'same-origin' }],
+      onData: [dispatchReceivedData, ['schedule']],
+      needed: false
+    },
```
mikx98: let's put this over the component displaying time, with `needed: true`

mikx98: since that component doesn't exist now, I approve. but please don't forget it

PR #351, opened 2020-02-19
Feature/add error message on not applied change

FrantisekMichalSebestyen:


src/actions/editPointsActions.js
```diff
@@ -0,0 +1,81 @@
...
+  return { newPoints, indices: [myIndex] }
+}
+
+const handleCopyWorklog = (points: Point[], indices: number[]) => {
+  return handleCreatePoint(
+    points,
+    points[indices[0]].description,
+    new Date().toISOString(),
+    points[indices[0]].issue
+  )
```
mikx98: this should recover selecting the last worklog

FrantisekMichalSebestyen: hm... we didn't want to import from ./trackerActions :(

mikx98: We don't need to, this function in editPoints will not be setting cursor. We decorate it here, add another decoration in trackerActions and there we also add the cursor setting. wdyt?

src/actions/trackerActions.js
```diff
@@ -337,11 +256,20 @@ const makeSelectionAction = actionCreator => (...args: any[]) => (
...
 }
 
+export { editTimestamp, createPoint }
+export const copySelectedWorklog = (...args: any[]) => (
+  dispatch: Dispatch,
+  getState: GetState
+) => {
+  const { points } = getState()
+  dispatch(makeSelectionAction(copyWorklog)(...args))
+  dispatch(setCursor(points.length - 1))
```
mikx98: I think the safest way is to write it as this:
```suggestion
  dispatch(setCursor(getState().points.length - 1))
```
because the copy worklog may insert one or not insert one (if nothing is selected nor hovered). also after https://github.com/vacuumlabs/fine-tracker/issues/354, it will be copying possibly multiple worklogs :)

src/tracker/ErrorBar/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+type Props = {|
+  errorMessage: ?string,
+  classes: Object,
+  handleClose: () => void
+|}
+
+const ErrorBar = ({ errorMessage, handleClose, classes }: Props) => (
+  <Snackbar
+    anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
```
mikx98: write this into the styles constant and write here something like:
```suggestion
    anchorOrigin={classes.anchorOrigin}
```

FrantisekMichalSebestyen: Is this actually doable? As I see it, styles are used to modify css and parts of theme. anchorOrigin is a prop.

mikx98: hm, depends on how the anchorOrigin works. maybe it can accept also a class name with the styles. please try and if it doesn't work, just resolve my comment :)

PR #355, opened 2020-02-27
changed kbd margin

### LINKED ISSUE #347: If the description is long, the end is lost under the shortcut helper
Screenshot from 2020-02-18 11-23-38

SurgeS:
I've moved it up a bit but this was already partially fixed after the new redesign.

image

image

resolves #347


FrantisekMichalSebestyen:
I think It would be better to change padding on Text field instead. wdyt ?

SurgeS:
> I think It would be better to change padding on Text field instead. wdyt ?
not sure...I did it this way mainly because: a) it's universal for all future input fields. b) the padding is static but the width of a shortcut hint changes based on its label.

You can check out the second commit, let me know what's better.

Issue #310, opened 2020-01-21
Not loggable worklog


jakuub:
Marking worklog as "not loggable" to be able to "assign" it to "free time" - it should work the same as empty description, but also should not contain information that it has no issue assigned (`NONE` on the place of issue key). Maybe we can add "FREE" or something like this. 

I suggest we can consult this with @darjanin . 

CC: @tamasbogi as a reporter of the issue.

mikx98:
@SurgeS did we consult with @darjanin ? if no, I think we should :)

PR #356, opened 2020-02-28
added free time issue

SurgeS:
added helper issues (free time issue for now) to use on FE

solves #310


src/consts.js
```diff
@@ -29,3 +29,8 @@ export const UNDOABLE_STATE = {
 
 export const SPEC_URL =
   'http://htmlpreview.github.io/?https://github.com/vacuumlabs/fine-tracker/blob/master/spec/dist/index.html'
+
+export const FREE_TIME = {
+  key: 'FREE',
```
SurgeS: I should probably add some internal key `prefix_jiraKey` to use on FE, since they need to be unique right?

